PR #9, opened 2016-05-23
Separate config reader to a new module. Fixes #8

### LINKED ISSUE #8: Make email credentials a bit secure
https://github.com/scorelab/Bassa/blob/master/components/core/EMail.py#L3

No hard coding accounts on the code :)

malithsen:


rehrumesh:
https://github.com/scorelab/Bassa/issues/8

rehrumesh:
Reviewed. Able to Merge. :+1:

components/core/email.conf
```diff
@@ -0,0 +1 @@
+{"username": "bassa.ucsc@gmail.com", "password": "bassa.123"}
```
manushanga: No. Nononono. Never ever store config within source, every major os gives you facilities to enable per user conf, even "~" resolvesv to user dir in windows too these days i guess please use them

components/core/DownloadDaemon.py
```diff
@@ -1,16 +1,17 @@
...
 from EMail import send_mail
 from DiskMan import *
+from ConfReader import conf_reader
 
 import websocket
 import sys
 
 
-conf = {}
+conf = conf_reader("dl.conf")
```
manushanga: Since this method returns an object how about get_conf_reader or create_conf_reader?

components/core/EMail.py
```diff
@@ -1,10 +1,11 @@
...
+from ConfReader import conf_reader
 
-username = 'bassa.ucsc@gmail.com'
-password = 'bassa.123'
+conf = conf_reader("email.conf")
 
 def send_mail(toaddrs, msg):
+    conf_reader()
     server = smtplib.SMTP_SSL('smtp.gmail.com:465')
-    server.login(username,password)
```
manushanga: Hmmm if you make the connection to a host, you are the client not the server :33

components/core/EMail.py
```diff
@@ -1,10 +1,11 @@
 import smtplib
+from ConfReader import conf_reader
 
-username = 'bassa.ucsc@gmail.com'
-password = 'bassa.123'
+conf = conf_reader("email.conf")
 
 def send_mail(toaddrs, msg):
+    conf_reader()
```
manushanga: Why read again? Without arguments am i missing something here?

components/core/ConfReader.py
```diff
@@ -0,0 +1,26 @@
...
+    emailConf = get_conf_reader('email.conf')
+
+    if any(val == '' for val in dlConf.values()):
+        sys.exit("Please set all the values in dl.conf")
+    if any(val == '' for val in emailConf.values()):
+        sys.exit("Please set all the values in email.conf")
+
+
+def get_conf_reader(confFile):
+    f = open(os.path.dirname(os.path.abspath(inspect.getfile(inspect.currentframe()))) + "/" + confFile)
```
manushanga: Still we are getting the conf file from source path right? Lets decide on a conf file dir like ~/.config/bassa on unix and %app_data%/bassa on windows ? And lets move the temp conf to that path upon install or from bassa itself

malithsen:
Updated. Tested only on Linux

components/core/email.conf
gnomezgrave: You mean to remove config files from Git? Yes, agree on that. However, place an example config file (i.e. email.conf.example) so that people can change it on their instances.

PR #15, opened 2016-06-28
View completed downloads

malithsen:
* Store file name and file size in db
* Show completed downloads
* Use angular-moment for relative time


ui/src/app/dashboard/controllers/DashCtrl.js
```diff
@@ -2,31 +2,25 @@
...
+    .controller('DashCtrl', [ '$scope', 'ToastService', 'DashService',
       DashCtrl
     ]);
 
-  function DashCtrl($scope, $http, ToastService) {
-    $scope.dlink = {link: ''}
+  function DashCtrl($scope, ToastService, DashService) {
+    $scope.dlink = {link: ''};
+    $scope.downloads = [];
     console.log(ToastService);
```
rehrumesh: Remove unnecessary console logs

ui/src/app/dashboard/services/DashService.js
```diff
@@ -0,0 +1,32 @@
...
+        .service('DashService', [
+        '$http',
+      dashService
+  ]);
+
+  function dashService($http){
+    var addDownload = function(link) {
+      return $http({
+          method: 'POST',
+          url: 'http://localhost:5000/api/download',
```
rehrumesh: Create an const value for the base url. May be an anglar value? And refer it here.
`url: BassaUrl+ 'api/download'`

easier to configure when moving to production. Or else will have to change this value in number of places.

ui/src/app/dashboard/services/DashService.js
```diff
@@ -0,0 +1,32 @@
...
+          data: JSON.stringify(link),
+          headers: {'Content-Type': 'application/x-www-form-urlencoded'}
+      });
+    };
+
+    var getDownloads = function() {
+      return $http({
+        method: 'GET',
+        url: 'http://localhost:5000/api/downloads/1'
```
rehrumesh: Shall we update the api and change download statuses into words instead of numbers? It will increase the readability don't you think @paarandika ?

malithsen: Number 1 in line 23 is the page number not a status. I've hard coded it for the time being. Will be fixed in the next PR.

PR #21, opened 2016-08-06
Data usage of top ten heaviest users #wip

malithsen:
This shows the usage of past month's top ten heaviest users in a pie chart along with the exact data usage.
Might have to alter this after user blocking feature is implemented.

### DIFF
--- a/ui/src/app/admin/controllers/AdminCtrl.js
+++ b/ui/src/app/admin/controllers/AdminCtrl.js
@@ -2,21 +2,12 @@
 
   angular
     .module('app')
-    .controller('AdminCtrl', [ '$scope', 'ToastService', 'AdminService', AdminCtrl]);
+    .controller('AdminCtrl', [ '$scope', 'ToastService', 'AdminService', 'UtilityService', AdminCtrl]);
 
   function AdminCtrl($scope, ToastService, AdminService) {
 
     $scope.signup_requests = [];
     $scope.usageChartData = [];
-
-    var formatBytes = function(bytes, decimals) {
-       if(bytes == 0) return '0 Byte';
-       var k = 1000;
-       var dm = decimals + 1 || 3;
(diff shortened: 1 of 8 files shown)


ui/src/app/admin/controllers/AdminCtrl.js
```diff
@@ -7,6 +7,36 @@
   function AdminCtrl($scope, ToastService, AdminService) {
 
     $scope.signup_requests = [];
+    $scope.usageChartData = [];
+
+    var formatBytes = function(bytes, decimals) {
+       if(bytes == 0) return '0 Byte';
+       var k = 1000;
```
rehrumesh: use following format. Also use a meaningful word instead of 'i'

```javascript
var k=1000,
    dm = decimals +1 ||3,
    sizes = [..... ],
    i = .... ;
```

ui/src/app/completed/controllers/TableCtrl.js
```diff
@@ -8,7 +8,24 @@
     $scope.dlink = {link: ''};
     $scope.downloads = [];
 
+    var formatBytes = function(bytes, decimals) {
+       if(bytes == 0) return '0 Byte';
+       var k = 1000;
```
rehrumesh: same as above comment

ui/src/app/completed/controllers/TableCtrl.js
```diff
@@ -8,7 +8,24 @@
     $scope.dlink = {link: ''};
     $scope.downloads = [];
 
+    var formatBytes = function(bytes, decimals) {
```
rehrumesh: I noticed this method has been declared in multiple places. Can't we create a common function?

malithsen: @rehrumesh How about a `UtilityService.js` ?

PR #58, opened 2017-01-11
Adding use strict to js files

lazypanda10117:
I added 'use strict'; to the js files according to this link: https://www.bithound.io/github/scorelab/Bassa/master/files#lint.


lazypanda10117:
updates: removed DS_STORE files.

Issue #102, opened 2017-03-20
Adding the option of deleting a download.


saisankargochhayat:
Implementing a delete a download from the ongoing download page , before it has been started . Also on the status of the download and its current progress. Is it a bug or still to be implemented ?
@rehrumesh @malithsen

saisankargochhayat:
Merged. #105

PR #105, opened 2017-03-22
[Feature] Remove download implemented in Dashboard

saisankargochhayat:
Solves #102.
Added a delete button in the dashboard.
image
Currently it works if the download id is less than 10.
If the id is 10 it fails . 
@malithsen if you could give a look.

### DIFF
--- a/components/core/DownloadManager.py
+++ b/components/core/DownloadManager.py
@@ -47,6 +47,7 @@ def remove_download(id, userName):
             cursor.execute(sql, (id, userName))
             db.commit()
         except MySQLdb.Error as e:
+            print(e)
             db.rollback()
             return e[1]
         return "success"
--- a/components/core/DownloadManager.py
+++ b/components/core/DownloadManager.py
@@ -111,7 +111,6 @@ def get_downloads_user(userName, limit):
 def get_downloads(limit):
     recordsPerPage = 15
     db = threadpool.connect()
-    print("hey i am here")
     if db is not None:
         cursor = db.cursor(MySQLdb.cursors.DictCursor)
         sql = "SELECT * FROM download WHERE status=3 ORDER by 'added_time' DESC LIMIT %s, %s;"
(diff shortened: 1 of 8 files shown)


ui/src/app/views/dashboard.html
```diff
@@ -16,14 +16,18 @@ <h4 class="panel-widget-tittle"> On going downloads</h4>
     <table id="table" class="table table-hover table-bordered">
         <thead>
         <tr ng-repeat="data in downloads">
-            <td data-title="ID">{{$index + 1}}</td>
-            <td data-title="Issue">{{data.download_name}}</td>
-            <td data-title="Progress">
+            <td style="padding-top: 3%" data-title="ID">{{$index + 1}}</td>
```
malithsen: Let's not use inline styles

components/core/DownloadManager.py
```diff
@@ -36,15 +36,15 @@ def remove_download(id, userName):
     db = threadpool.connect()
     if db is not None:
         cursor = db.cursor()
-        sql1 = "SELECT status FROM download WHERE id=%s;"
+        sql1 ="SELECT status FROM download WHERE id=%s;"
         sql = "DELETE from download WHERE id=%s and user_name=%s;"
         try:
-            cursor.execute(sql1, id)
+            cursor.execute(sql1,str(id))
```
saisankargochhayat: If the value of id exceeds 10 , delete fails. Please suggest something regarding this.

ui/src/app/views/dashboard.html
saisankargochhayat: Will remove them :)

saisankargochhayat:
The issue #102 has been fixed . Currently works for all id values . Please review :)

ui/src/app/stylesheets/_table.scss
```diff
@@ -210,3 +210,7 @@ $table-border-color:            #e0e0e0;
     }
   }
 }
+
+.vertalign{
+  padding-top: 3% !important;
```
malithsen: Make it an id and you can get rid of the `!important`. Also IMO _custom.scss is a better place to have this.

saisankargochhayat:
Added a confirmation dialogue . :smile: @malithsen
pr

Issue #117, opened 2017-04-03
Downloads in the Bassa server fails


rehrumesh:
When bassa try to download multiple files, it crashes the app,

## Expected Behavior
Bassa should be able to handle concurrent downloads without any issues.

## Current Behavior
If there are multiple downloads, server crashes.

## Possible Solution
Remove the stat module.

saisankargochhayat:
I will work on this , to find the exact cause , then fix it . @rehrumesh

saisankargochhayat:
On increasing the threadpool and updating SQLAlchemy versions , the multiple download seems to work fine. Will try this a few more times and make a PR .

vivonk:
I guess the issue is resolved and download is even working fine on the queue of 20+ links. 
Closing the issue. Further, if anyone wants to reopen, the issue would be around the same thing.

PR #120, opened 2017-04-13
Multiple downloads crash

saisankargochhayat:
Fixing issue #117 
## Description
Increased threadpool.
Updated SQLAlchemy

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,7 @@
 
 *.log
 components/core/debug-server.log
+components/core/email.conf
 downloads/
 
 # UI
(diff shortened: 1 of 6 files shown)


malithsen:
Good job! Will merge after testing locally :1st_place_medal: :confetti_ball:

paarandika:
Can we add this type of configurations to a global conf file fetch them from there? Unlike in the beginning of the project, now there are a number of configurations scattered in the code and it's hard to keep track of them.
It would be better if we can address this with filing a separate issue. Just mentioned this here since these configs are really important.

saisankargochhayat:
Yes indeed , that would be a good way to keep track of them , we could make a list of what other config's we need in the config file and add it as a global config file.

Issue #125, opened 2017-04-28
[Feature] Current user's name is not displayed on Logout menu 


KaveeshBaddage:
## Expected Behavior

When system user is going to logout from the system, current user's user name should view in the Logout menu by the system.

## Current Behavior

There is only logout button in the logout menu. That doesn't displayed who is the current user.

![bassalogoutview](https://cloud.githubusercontent.com/assets/17790688/25515749/4dbc02da-2c04-11e7-9e28-8aaeae3c460a.png)

## Possible Solution

Adding current user's user name in to the logout menu.

![bassalogoutmenuwithusername](https://cloud.githubusercontent.com/assets/17790688/25515780/67a988b6-2c04-11e7-86ef-88ba3d15d1b5.png)

## Context

This gives the opportunity to the user to know from which account he has been used to login to the system.

## Your Environment

* Environment name and version :
       Firefox Version 53.0
       Node.js v6.10.0
 * Operating System and version :
        ubuntu 16.04 LTS
* Link to your project:
https://github.com/KaveeshBaddage/Bassa

PR #127, opened 2017-04-28
View current user's user name on logout menu. Resolving Issue #125

KaveeshBaddage:
## Description

When system user is going to logout from the system, current user's user name should view in the Logout menu by the system.Currently  that feature is not included. This pull request add required implementation to view current user's name in the logout menu.
## Motivation and Context

This gives the opportunity to the user to know from which account he has been used to login to the system.

## How Has This Been Tested?

After implementing this feature, I have logged to the system using different user accounts and checked is it work correctly. It's worked correctly all the time.

## Screenshots :
View of the logout menu before implementing

bassalogoutview

View of the logout menu after implementing

bassalogoutmenuwithusername

## Types of changes

- [ #] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to hange)

## Checklist:

- [ #] My code follows the code style of this project.

### DIFF
--- a/ui/src/app/login/controllers/LoginCtrl.js
+++ b/ui/src/app/login/controllers/LoginCtrl.js
@@ -6,9 +6,6 @@
 
   function LoginCtrl($scope, $state, UserService) {
     $scope.user = {};
- 
-
-
     $scope.login = function(){
       $scope.incorrectCredentials = false;
       UserService.login($scope.user, function(status) {
--- a/ui/src/app/login/controllers/LoginCtrl.js
+++ b/ui/src/app/login/controllers/LoginCtrl.js
@@ -6,15 +6,13 @@
 
   function LoginCtrl($scope, $state, UserService) {
     $scope.user = {};
-    //$rootScope.usernme = $scope.user.user_name;;
-    console.log("hello" + $scope.user.user_name);
+ 
 
 
     $scope.login = function(){
       $scope.incorrectCredentials = false;
       UserService.login($scope.user, function(status) {
         if (status){
-          console.log("user: " + JSON.stringify($scope.user));
           localStorage.setItem("user", JSON.stringify($scope.user));
           $state.go('home.dashboard');
         } else {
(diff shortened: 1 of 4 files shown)


rajikaimal:
Codacy is complaining about not having `camel_case` notation in `MainController.js`.

KaveeshBaddage:
In Maincontroller.js file, I have used a variable named "user_name". That variable was declared in login.html file. It was existed variable in login.html when I was  starting the implementation of this functionality and I didn't go th change that variable. Does it need to change from the beginning?

rajikaimal:
If that's the case changing `user_name` might cause breaking changes. Better to ping @malithsen about this. :running_man:

KaveeshBaddage:
@rajikaimal Ok cool..  @malithsen Does it need to change "user_name"  variable name from the beginning or Is it okay  to use in this form ?

malithsen:
@KaveeshBaddage Nice work :+1: 
I'll merge the PR as is.
Please consider issuing another PR addressing the styling inconsistencies
1) `user_name` 
2) We have used both `var vm = this` and $scope. Since we have used $scope in most files, I think it's better to stay with it. So let's replace `vm` with `$scope` in `MainController.js` as well

ui/src/app/login/controllers/LoginCtrl.js
```diff
@@ -6,11 +6,14 @@
 
   function LoginCtrl($scope, $state, UserService) {
     $scope.user = {};
+
```
malithsen: Let's remove these empty lines

KaveeshBaddage: Okay. I will work on that problem.

KaveeshBaddage:
Thank you @malithsen. :smile:  I have removed empty lines from LoginCtrl.js file. I will work on those issues which you notified above.

malithsen:
https://github.com/scorelab/Bassa/blob/9d1bdd666584e5a63461464a2528e9962974528f/ui/src/app/login/controllers/LoginCtrl.js#L13

Just noticed that this stores the password in plaintext in localstorage :scream:

PR #129, opened 2017-05-06
Replace variable vm with $scope variable

KaveeshBaddage:
## Description

Both var vm = this and $scope had been used in the MainController.js and main.html files. Since we have used $scope in most files, I think it's better to stay with it. So I replace vm with $scope in MainController.js and main.html files.

## Related Issue

This issue is suggested by @malithsen while working with the 127th pull request( pull request - View current user's username on logout menu. Resolving Issue #125)
 
## Motivation and Context

Since we have used $scope in most files, I think it's better to stay with it

## How Has This Been Tested?

I have logged to the system using different user accounts and checked is it work correctly. It's worked correctly all the time.

## Types of changes

- [x ] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [ x] My code follows the code style of this project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,12 +58,12 @@ mysql -u root -p  Bassa < Bassa.sql
 
 ## URL endpoints
 
-**http://localhost:5000/api/login**
+**/api/login**
 
 Form data: user_name, password
-Returns auth token in response header for successful login
+Returns a auth token in response header for a successful login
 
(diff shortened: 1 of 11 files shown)


KaveeshBaddage:
A new pull request is going to add using these implemented functionalities.

PR #243, opened 2017-12-10
Dockerize Bassa Core

kmehant:
@agentmilindu Updated the dockerfile as per the given instructions in the GCI task. Kindly Review.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,14 +1,15 @@
 FROM python:3
 MAINTAINER SCoRe Lab Community <commuity@scorelab.org>
 
-RUN apt-get update
+WORKDIR /home/Bassa
 
-RUN mkdir -p /home/Bassa
-RUN apt-get install git
-RUN git clone https://github.com/scorelab/Bassa.git /home/Bassa
-RUN chmod 755 /home/Bassa/setup.sh
-RUN /home/Bassa/setup.sh
-RUN python components/core/setup.py develop
+COPY setup.sh  . \
+&&   components/core components/core
+
+RUN apt-get update \
+&&  chmod 755 setup.sh && \
+&&  setup.sh && \
+&&  python components/core/setup.py develop
 
 CMD [ "python", "components/core/Main.py" ]
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,13 +3,12 @@ MAINTAINER SCoRe Lab Community <commuity@scorelab.org>
 
 WORKDIR /home/Bassa
 
-COPY setup.sh  . \
-&&   components/core components/core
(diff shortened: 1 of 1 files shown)


kmehant:
@agentmilindu All tests passed :) Kindly review.

Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM python:3
+MAINTAINER SCoRe Lab Community <commuity@scorelab.org>
+
+RUN apt-get update
+
+RUN mkdir -p /home/Bassa
+RUN apt-get install git
+RUN git clone https://github.com/scorelab/Bassa.git /home/Bassa
```
agentmilindu: Hi @kmehant, we don't need cloning the repo with git here. We can simply copy the files from the local repo. 

You can refer: https://docs.docker.com/engine/reference/builder/#copy

Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM python:3
+MAINTAINER SCoRe Lab Community <commuity@scorelab.org>
+
+RUN apt-get update
+
+RUN mkdir -p /home/Bassa
```
agentmilindu: You can set `/home/Bassa` as the `WORKDIR`, then you can simply say, 

`RUN chmod 755 setup.sh`
`RUN setup.sh`

Refer: https://docs.docker.com/engine/reference/builder/#workdir

Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+RUN apt-get install git
+RUN git clone https://github.com/scorelab/Bassa.git /home/Bassa
+RUN chmod 755 /home/Bassa/setup.sh
+RUN /home/Bassa/setup.sh
```
agentmilindu: One good practice writing dockerfiles is combining the RUN commands together if they relate. 

You can do as follows, 

```
RUN apt-get update \
&&  chmod 755 /home/Bassa/setup.sh && \
&&  /home/Bassa/setup.sh && \
&&  python components/core/setup.py develop
```

Refer this : https://hackernoon.com/tips-to-reduce-docker-image-sizes-876095da3b34

Dockerfile
kmehant: @agentmilindu I have a doubt here, can i say

```
WORKDIR /home/Bassa
COPY Bassa WORKDIR 
```

Is this valid?

agentmilindu: You can say `COPY . /home/Bassa `

http://blog.cloud66.com/how-to-get-code-into-a-docker-container/#option3usingtheaddorcopycommand

kmehant:
@agentmilindu Thank you for the guidance, i addressed all the changes :)

PR #246, opened 2017-12-10
Correct the extra && in Dockerfile

kmehant:
I made a small mistake in my previous pull request, hence sending this one to correct it.


kmehant:
This is fixed as part of #247

PR #249, opened 2017-12-10
Create docker-compose.yml

kmehant:
Fixes #249

### DIFF
--- a/components/core/Dockerfile
+++ b/components/core/Dockerfile
@@ -1,0 +1,23 @@
+FROM python:3
+
+MAINTAINER SCoRe Lab Community <commuity@scorelab.org>
+
+WORKDIR /home/Bassa
+
+COPY setup.sh .
+COPY components/core components/core
+
+RUN apt-get update \
+&& ./setup.sh \
+&& python components/core/setup.py develop 
+
+CMD [ "python", "components/core/Main.py" ]
+
+LABEL multi.org.label-schema.name = "Bassa" \
+      multi.org.label-schema.description = "Bassa provides Automated Download Queue to make the best use of Internet bandwidth" \
+      multi.org.label-schema.url="https://github.com/scorelab/Bassa/wiki" \
+      multi.org.label-schema.vcs-url = "https://github.com/scorelab/Bassa" \
+      multi.org.label-schema.vcs-ref = "" \
+      multi.org.label-schema.vendor = "Sustainable Computing Research Group" \
+      multi.org.label-schema.version = "" \
+      multi.org.label-schema.schema-version = "1.0"
--- a/ui/Dockefile
+++ b/ui/Dockefile
@@ -1,16 +1,15 @@
-FROM python:3
+FROM node
+
 MAINTAINER SCoRe Lab Community <commuity@scorelab.org>
 
 WORKDIR /home/Bassa
 
-COPY setup.sh .
-COPY components/core components/core
+COPY ui .
 
 RUN apt-get update \
-&& ./setup.sh \
-&& python components/core/setup.py develop
+&& npm install
 
-CMD [ "python", "components/core/Main.py" ]
+CMD ["gulp", "serve"]
 
 LABEL multi.org.label-schema.name = "Bassa" \
       multi.org.label-schema.description = "Bassa provides Automated Download Queue to make the best use of Internet bandwidth" \


docker-compose.yml
```diff
@@ -0,0 +1,25 @@
+version: "3"
+services:
+  web:
+    build:
+      context: ui/
+      dockerfile: Dockerfile
```
agentmilindu: This line is not needed

docker-compose.yml
```diff
@@ -0,0 +1,25 @@
...
+      ports:
+        - "3000:3000"
+      labels:
+        com.example.description: "Docker container for Bassa UI"
+  api:
+    build:
+      context: components/core
+      dockerfile: Dockerfile
```
agentmilindu: This line is not needed.

docker-compose.yml
```diff
@@ -0,0 +1,25 @@
...
+        - "8000:8000"
+      labels:
+        com.example.description: "Docker container for Bassa backend"
+  db:
+    image: mysql
+    restart: always
+    ports:
+      - 6603:3306
+    environment:
+      MYSQL_ROOT_PASSWORD: ${BASSA_DB_PASSWORD}
```
agentmilindu: Let's add `BASSA_DB_USERNAME` elso.

docker-compose.yml
```diff
@@ -0,0 +1,25 @@
+version: "3"
+services:
+  web:
+    build:
+      context: ui/
```
agentmilindu: Let's remove both `context` and `dockerfile` and keep the `build`
```
web:
  build: .
```

docker-compose.yml
```diff
@@ -0,0 +1,25 @@
...
+      dockerfile: Dockerfile
+      ports:
+        - "3000:3000"
+      labels:
+        com.example.description: "Docker container for Bassa UI"
+  api:
+    build:
+      context: components/core
```
agentmilindu: Let's remove both `context` and `dockerfile` and keep the `build`
```
web:
  build: .
```

docker-compose.yml
```diff
@@ -0,0 +1,25 @@
...
+      dockerfile: Dockerfile
+      ports:
+        - "8000:8000"
+      labels:
+        com.example.description: "Docker container for Bassa backend"
+  db:
+    image: mysql
+    restart: always
+    ports:
+      - 6603:3306
```
agentmilindu: Any specific reason to change the port?

docker-compose.yml
```diff
@@ -0,0 +1,25 @@
...
+        - "3000:3000"
+      labels:
+        com.example.description: "Docker container for Bassa UI"
+  api:
+    build:
+      context: components/core
+      dockerfile: Dockerfile
+      ports:
+        - "8000:8000"
+      labels:
```
agentmilindu: Let's add the following part also,
```
depends_on:
      - "db"
```

docker-compose.yml
agentmilindu: Let's do as following, 

```
MYSQL_DATABASE: ${BASSA_DB_NAME}
MYSQL_USER: ${BASSA_DB_USERNAME}
MYSQL_PASSWORD: ${BASSA_DB_PASSWORD} 
```

and remove `MYSQL_ROOT_PASSWORD: ${BASSA_DB_PASSWORD}`

docker-compose.yml
```diff
@@ -0,0 +1,25 @@
...
+        com.example.description: "Docker container for Bassa backend"
+  db:
+    image: mysql
+    restart: always
```
agentmilindu: We need to set volumes also, 
```
volumes:
      - data-volume:/var/lib/mysql
```

agentmilindu:
At the end of the file, add the following, with the service's level,

```
volumes:
  data-volume:
```

Refer: https://docs.docker.com/compose/compose-file/#volume-configuration-reference

kmehant:
@agentmilindu addressed the changes :) kindly review

components/core/setup.sh
```diff
@@ -0,0 +1,34 @@
...
+echo "Installing mysql-server"
+apt-get install -y mysql-server
+echo "mysql-server installed"
+
+echo "Installing libmysqlclient"
+apt-get install libmysqlclient-dev
+echo "libmysqlclient installed"
+
+echo "Installing node"
+curl -sL https://deb.nodesource.com/setup_6.x | sudo -E bash -
```
agentmilindu: Let's remove the `sudo` here also. 

eg: `curl -sL https://deb.nodesource.com/setup_6.x | bash -`

PR #318, opened 2017-12-23
ESLint for Bassa UI

kmehant:


kmehant:
@agentmilindu

ui/.eslintrc.yml
```diff
@@ -0,0 +1,6 @@
+extends: airbnb-base
+plugins:
+    - import
+rules:
+    no-underscore-dangle: [2, {allowAfterThis: true}]
+    linebreak-style: [error, windows]
```
agentmilindu: I'm not sure if this `linebreak-style: windows ` would give Linux users troubles.

kmehant: shall i remove this rule?

agentmilindu: @rehrumesh @malithsen WDYT? I'm not very familiar with ESLint

PR #323, opened 2017-12-23
Create production Docker for Bassa UI with Nginx

kmehant:


ui/Dockerfile
```diff
@@ -1,20 +1,14 @@
...
 
-COPY . .
-
-RUN apt-get update \
-&& npm install -g gulp \
-&& npm install -g bower \
-&& npm install --unsafe-perm
-
-CMD ["gulp", "serve"]
+RUN apt-get update
```
agentmilindu: Let's get this line before the COPY lines. Would be easy for layer caching.

kmehant:
@agentmilindu

ui/Dockerfile
```diff
@@ -1,20 +1,14 @@
-FROM node
+FROM nginx:alpine
 
 MAINTAINER SCoRe Lab Community <commuity@scorelab.org>
 
 ARG BUILD_DATE
 ARG VCS_REF
 
-WORKDIR /home/Bassa
+RUN apt-get update
```
agentmilindu: Alpine Linux does not use `apt-get` it seems, we have to replace this with an equivalent.

ui/Dockerfile
```diff
@@ -1,20 +1,14 @@
...
-COPY . .
-
-RUN apt-get update \
-&& npm install -g gulp \
-&& npm install -g bower \
-&& npm install --unsafe-perm
-
-CMD ["gulp", "serve"]
+COPY nginx.conf /etc/nginx/conf.d/bassa.conf
+COPY ui/dist /var/www/bassa
```
agentmilindu: Should be `COPY dist /var/www/bassa`

PR #328, opened 2017-12-24
Add shell script to echo distro and database client preference

padamchopra:
## Add shell script to detect and echo the linux distro and ask the user of their preferred sql client
[GCI Task]

**Results of Script**
bashscriptupdated

### DIFF
--- a/scripts/setup-database.sh
+++ b/scripts/setup-database.sh
@@ -1,25 +1,33 @@
 #!/bin/bash
 distrofull=$(cat /etc/issue);
 distro=${distrofull:: -6};
-echo "Enter number of the Database Client you prefer:";
-echo "1. MySQL";
-echo "2. SQLite";
-echo "3. MariaDB";
-echo "4. PostgreSQL";
-read user_preference;
+echo "Select the Database Client you prefer:";
+options=("MySQL" "SQLite" "MariaDB" "PostgreSQL");
+select opt in "${options[@]}"
+do
+    case $opt in
+        "MySQL")
+            user_preference="MySQL"
+	    break
+            ;;
+        "SQLite")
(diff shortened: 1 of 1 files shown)


scripts/setup-database.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+distrofull=$(cat /etc/issue);
+distro=${distrofull:: -6};
+echo "Enter number of the Database Client you prefer:";
```
agentmilindu: See this answer, it looks tidy https://askubuntu.com/a/1716/740820
Shall we try in that manner?

scripts/setup-database.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+distrofull=$(cat /etc/issue);
```
agentmilindu: See https://unix.stackexchange.com/a/84375
Let's have several  options in hand and test on all the distros we support. 

One more methos is here. 
```
cat /etc/*-release | grep "DISTRIB_ID=" | cut -d"=" -f2
```

padamchopra: I have two distros on my system, works good on both. So maybe we can try going with this first :worried:

malithsen: `cat /etc/*-release | grep "DISTRIB_ID=" | cut -d"=" -f2` is better IMO. 
1. `${distrofull:: -6};` this might not always slice out all the escape characters
2. It's better to detect the base distro. Arch in this case. Which is correctly identified by the other command
discheck

padamchopra: DONE

padamchopra:
@malithsen The requested Change has been made. Thanks for the better code suggestion :)

PR #340, opened 2017-12-25
Add setup.sh

padamchopra:
## Add setup.sh in scripts directory
[GCI 2017]

I've made the script according to the conditions stated in the task-
```
Create a directory called scripts/debian/ and create a file as setup.sh there also, 
and add echo "Setting up Bassa on Debian bases Linux".

scripts/setup.sh should do the following
1) Check if the terminal has root permission. If not exit the terminal after showing
   message asking the user to run with root privileges
2) Find the Linux distribution of the current machine and store it in a variable
3) Check if the Linux distribution of the machine is debian, if so execute the 
   file located in scripts/debian/setup.sh
```

**Working of script in all scenarios:**
bassasetupsh

### DIFF
--- a/scripts/debian/setup.sh
+++ b/scripts/debian/setup.sh
@@ -1,2 +1,2 @@
 #!/bin/bash
-echo "Setting up Bassa on Debian bases Linux";
+echo "Setting up Bassa on Debian based Linux";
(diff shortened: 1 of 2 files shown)


scripts/debian/setup.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+echo "Setting up Bassa on Debian bases Linux";
```
agentmilindu: "Setting up Bassa on Debian based Linux"

kmehant:
As we have now got a dedicated dependency installation shell script across various operating systems (Except Windows OS). We shall close this PR for now.

PR #345, opened 2017-12-26
Redesign Bassa Signup Page

kmehant:
screen shot 1939-10-05 at 7 36 28 pm
screen shot 1939-10-05 at 7 36 39 pm
screen shot 1939-10-05 at 7 36 52 pm

The login and signup buttons will be activated only when the angular form validations are passed. We give the validations at form level itself before even submitting the form for validating at controller level.

### DIFF
--- a/ui/src/app/stylesheets/_custom.scss
+++ b/ui/src/app/stylesheets/_custom.scss
@@ -180,49 +180,6 @@ md-backdrop.md-sidenav-backdrop {
   margin-top: 0px;
 }
 
-#signup-button{
-  border-color: #757575;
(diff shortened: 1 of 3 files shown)


malithsen:
Hi @kmehant,
Please add screenshots of the changes you've made

malithsen:
Hi @kmehant,
* Let's keep the login page as it was.

* "Username should be more than 5 characters and less than 10" 
Let's make it 3 and 15

* Make the buttons center aligned

* Emphasis should be on the back button, instead of the signup button

* Let's change it to "Sign Ip for Bassa" and make it center aligned

PR #353, opened 2017-12-27
Python script to setup db

padamchopra:
## Added init-database.py to setup database
[GCI 2017]

The script communicates with sql server using sqlalchemy module for python.

Screenshot for the database setup by python script
bassapythonscript

### DIFF
--- a/scripts/init-database.py
+++ b/scripts/init-database.py
@@ -2,42 +2,49 @@
 
 import yaml
 import sqlalchemy
-import MySQLdb
+import os
 
-values = list()
+values = 'configurations'
+path = 'root_path'
 
 
-def retreiveValues():
-    stream = open("../.config/bassa/config.yml", "r")
-    docs = yaml.load_all(stream)
+def retreive_values():
+    global path
+    path= os.path.abspath(os.path.join(os.path.join(__file__,os.pardir),os.pardir))
+    stream = open(path+"/bassa.yml", "r")
+    configs = yaml.load(stream)
     global values
-    for doc in docs:
-        for k, v in doc.items():
-            for m, n in v.items():
-                values.append(n)
+    values = configs
 
 
-def createDatabase():
-    param = values[5]+'://'+values[0]+':'+values[4]+'@'+values[2]
-    engine = sqlalchemy.create_engine(param)
-    engine.execute("CREATE DATABASE IF NOT EXISTS " + values[3])
+def create_database():
+    param = values['database']['database_type']+'://'+values['database']['database_user_username']+':'+values['database']['database_user_password']+'@'+values['database']['database_ip']
+    try:
+        engine = sqlalchemy.create_engine(param)
+	engine.execute("CREATE DATABASE IF NOT EXISTS " + values['database']['database_name'])
+    except:
+        param = values['database']['database_type']+'://root:'+values['database']['database_user_password']+'@'+values['database']['database_ip']
+	engine = sqlalchemy.create_engine(param)
+	engine.execute("CREATE DATABASE IF NOT EXISTS " + values['database']['database_name'])
+	engine.execute("GRANT ALL PRIVILEGES ON "+values['database']['database_name']+" TO "+values['database']['database_user_username']+"@"+values['database']['database_ip']+"WITH GRANT OPTION;")
 
 
-def importSQL():
-    param = values[5]+'://'+values[0]+':'+values[4]+'@'+values[2]+'/'+values[3]
-    engine = sqlalchemy.create_engine(param)
-    fd = open("../Bassa.sql", 'r')
-    sqlFile = fd.read()
+def import_SQL():
+    param = values['database']['database_type']+'://'+values['database']['database_user_username']+':'+values['database']['database_user_password']+'@'+values['database']['database_ip']+'/'+values['database']['database_name']
+    fd = open(path+"/Bassa.sql", 'r')
+    sql_File = fd.read()
     fd.close()
-    sqlCommands = sqlFile.split(';')
+    sqlCommands = sql_File.split(';')
     for command in sqlCommands:
         try:
             with engine.connect() as con:
                 con.execute(command+';')
         except:
-            error = 1
-retreiveValues()
-createDatabase()
-importSQL()
-print "Database (" + values[3] + ") is now successfuly setup"
+            pass
 
+
+if __name__ == "__main__":
+    retreive_values()
+    create_database()
+    import_SQL()
+    print "Database (" + values['database']['database_name'] + ") is now successfuly setup"
--- a/scripts/init-database.py
+++ b/scripts/init-database.py
@@ -3,6 +3,7 @@
 import yaml
 import sqlalchemy
 import os
+import getpass
 
 configs = 'configurations'
 path = 'root_path'
@@ -17,7 +18,8 @@ def retreive_values():
 
 
 def create_database():
-    param = configs['database']['database_type']+'://root:'+configs['database']['database_user_password']+'@'+configs['database']['database_ip']
+    root_password = getpass.getpass(prompt = 'Enter root password:')
+    param = configs['database']['database_type']+'://root:'+root_password+'@'+configs['database']['database_ip']
     engine = sqlalchemy.create_engine(param)
     engine.execute("CREATE DATABASE IF NOT EXISTS " + configs['database']['database_name'])
     engine.execute("CREATE USER "+configs['database']['database_user_username']+"@"+configs['database']['database_ip']+" IDENTIFIED BY " + configs['database']['database_user_password'])
--- a/scripts/init-database.py
+++ b/scripts/init-database.py
@@ -5,8 +5,8 @@ import sqlalchemy
 import os
 import getpass
 
-configs = 'configurations'
-path = 'root_path'
+configs = ''
+path = ''
 
 
 def retreive_values():
@@ -18,9 +18,8 @@ def retreive_values():
 
 
 def create_database():
-    try:
-        os.environ['MYSQL_ROOT_PASSWORD']
-    except:
+    root_password = os.environ['MYSQL_ROOT_PASSWORD']
+    if not root_password:
         root_password = getpass.getpass(prompt = 'Enter root password:')
         os.environ['MYSQL_ROOT_PASSWORD'] = root_password
     connection_url = configs['database']['database_type']+'://root:'+os.environ['MYSQL_ROOT_PASSWORD']+'@'+configs['database']['database_ip']
@@ -28,7 +27,6 @@ def create_database():
     engine.execute("CREATE DATABASE IF NOT EXISTS " + configs['database']['database_name'])
     engine.execute("CREATE USER "+configs['database']['database_user_username']+"@"+configs['database']['database_ip']+" IDENTIFIED BY " + configs['database']['database_user_password'])
     engine.execute("GRANT INSERT, UPDATE, SELECT, DELETE ON "+configs['database']['database_name']+".* TO "+configs['database']['database_user_username']+"@"+configs['database']['database_ip'])
-    engine.execute("FLUSH PRIVILEGES")
 
 
 def import_SQL():
--- a/scripts/init-database.py
+++ b/scripts/init-database.py
@@ -1,0 +1,28 @@
+#!/usr/bin/env python
+
+import yaml, sqlalchemy, MySQLdb
+
+stream = open("../.config/bassa/config.yml", "r")
+docs = yaml.load_all(stream)
+values = list()
+for doc in docs:
+    for k,v in doc.items():
+        for m,n in v.items():
+		values.append(n)
+dbtype=values[5]
+dbusername=values[0]
+dbpass=values[4]
+ip=values[2]
+# db=MySQLdb.connect(host="localhost", user="bassaman",passwd="s57d46857f968t79pho")
+# db1=db.cursor()
+# db1.execute('Create database test100')
+engine = sqlalchemy.create_engine(dbtype+'://'+dbusername+':'+dbpass+'@'+ip)
+engine.execute("CREATE DATABASE IF NOT EXISTS " + values[3])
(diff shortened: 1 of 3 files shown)


iammosespaulr:
I'm sorry but the task was such that the script must be able to read from the .yml file and be able to write to the SQL database, right ???

padamchopra:
Yes. And that's exactly what the script does.

iammosespaulr:
oh ?? could you select * from user ;
and post the output

malithsen:
@padamchopra Good job with the script :+1: 
Few suggestions
* Make the required alterations to follow pep8 style guide. You can use an online tool like http://pep8online.com/checkresult

* Separate the code into sensible functions to improve readability
eg: read_config(), dbcon() etc...

padamchopra:
@iammosespaulr Here you go.
user
@malithsen Ok, I'll work on the changes

iammosespaulr:
@padamchopra Nice   .... :+1:  ive been working on that task myself until i abandoned it , will send you a pm in gitter with the script i wrote . 
Regards.

padamchopra:
@malithsen I've made the changes. 
Also, I've made alterations to follow pep8 style guide
pep8

scripts/init-database.py
```diff
@@ -0,0 +1,43 @@
...
+
+import yaml
+import sqlalchemy
+import MySQLdb
+
+values = list()
+
+
+def retreiveValues():
+    stream = open("../.config/bassa/config.yml", "r")
```
agentmilindu: I think we should find the path of the file and then access the parent dir for this programmatically. 

We can use following for that, 
```
import os
path = os.path.abspath(__file__)
```
See https://stackoverflow.com/a/595315/1190184

Think what will happen if someone ran this script like `python scripts/init-database.py`? What will be the `../` folder be? Mu guess is, it will be the parent dir of the Bassa project dir. 

Anyway, the config file's path should be `bassa.yml` not `.config/bassa/config.yml`

scripts/init-database.py
```diff
@@ -0,0 +1,43 @@
...
+    docs = yaml.load_all(stream)
+    global values
+    for doc in docs:
```
agentmilindu: Why we collect all variables like this? Can't we access them only when needed, like `docs.database.database_name`

Anyway, `docs` is not a good name for this, shall we change it to `configs`?

scripts/init-database.py
```diff
@@ -0,0 +1,42 @@
...
+        for k, v in doc.items():
+            for m, n in v.items():
+                values.append(n)
+
+
+def createDatabase():
```
agentmilindu: @malithsen @padamchopra  We are using camel cases or snake cases for method names?

scripts/init-database.py
```diff
@@ -0,0 +1,42 @@
...
+    sqlFile = fd.read()
+    fd.close()
+    sqlCommands = sqlFile.split(';')
+    for command in sqlCommands:
+        try:
+            with engine.connect() as con:
+                con.execute(command+';')
+        except:
+            pass
+retreiveValues()
```
agentmilindu: Keep some space before this block. How about we put these inside a Python `__main__` check

```
if __name__ == "__main__":
   createDatabase()
   importSQL()
   print "Database (" + values[3] + ") is now successfuly setup"
```

scripts/init-database.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def createDatabase():
+    param = values[5]+'://'+values[0]+':'+values[4]+'@'+values[2]
+    engine = sqlalchemy.create_engine(param)
+    engine.execute("CREATE DATABASE IF NOT EXISTS " + values[3])
+
+
+def importSQL():
+    param = values[5]+'://'+values[0]+':'+values[4]+'@'+values[2]+'/'+values[3]
```
agentmilindu: Can't we use as `docs.database.database_name`?

Ideally should be like `configs.database.database_name`

scripts/init-database.py
```diff
@@ -0,0 +1,42 @@
...
+    engine = sqlalchemy.create_engine(param)
+    fd = open("../Bassa.sql", 'r')
```
agentmilindu: Same for this about getting the path. 

And the purpose of this script is to create a new database with the given db name if it does not exist, create a user with the with the given username and password if it does not exist and then grant the correct access right to the user to the Bassa database.

Finally, we can import the Bassa.sql.

scripts/init-database.py
```diff
@@ -0,0 +1,42 @@
...
+    sqlFile = fd.read()
+    fd.close()
+    sqlCommands = sqlFile.split(';')
+    for command in sqlCommands:
```
agentmilindu: Can't we run the whole file at once?

padamchopra:
@agentmilindu I've made all the suggested changes. Except for the one in which you've asked if the entire sql file can be run at once. I've been trying since morning and have got no luck. But I did try another approach in which I look at lines and if they don't have a ";" , I concatenate them to a string and if they do have a ";" I execute it as a command. But got no luck with it, you can have a look at it though- https://ideone.com/XKQB7m

scripts/init-database.py
malithsen: snake case

scripts/init-database.py
```diff
@@ -0,0 +1,56 @@
...
+path = 'root_path'
+
+
+def retreive_values():
+    global path
+    path= os.path.abspath(os.path.join(os.path.join(__file__,os.pardir),os.pardir))
+    stream = open(path+"/bassa.yml", "r")
+    configs = yaml.load(stream)
+    global values
+    values = configs
```
agentmilindu: Why assign `configs` to `values`? Can't we use `configs` directly?

scripts/init-database.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+def create_database():
+    param = values['database']['database_type']+'://'+values['database']['database_user_username']+':'+values['database']['database_user_password']+'@'+values['database']['database_ip']
```
agentmilindu: Let's change `param` to `connection_url`

agentmilindu: Can't we use `configs` directly? 

```
configs['database']['database_type']
```

scripts/init-database.py
```diff
@@ -0,0 +1,56 @@
...
+    try:
+        engine = sqlalchemy.create_engine(param)
+	engine.execute("CREATE DATABASE IF NOT EXISTS " + values['database']['database_name'])
+    except:
+        param = values['database']['database_type']+'://root:'+values['database']['database_user_password']+'@'+values['database']['database_ip']
+	engine = sqlalchemy.create_engine(param)
+	engine.execute("CREATE DATABASE IF NOT EXISTS " + values['database']['database_name'])
+	engine.execute("GRANT ALL PRIVILEGES ON "+values['database']['database_name']+" TO "+values['database']['database_user_username']+"@"+values['database']['database_ip']+"WITH GRANT OPTION;")
```
agentmilindu: @padamchopra @malithsen We have to decide what are the privileges this new user should have?

I guess that is just, `INSERT`, `UPDATE`, `SELECT` and `DELETE`. If we move to user use soft deletes, we can drop `DELETE` also.

scripts/init-database.py
```diff
@@ -0,0 +1,56 @@
...
+    sqlCommands = sql_File.split(';')
+    for command in sqlCommands:
+        try:
+            with engine.connect() as con:
+                con.execute(command+';')
+        except:
+            pass
+
+
+retreiveValues()
```
agentmilindu: Why we have these again whence we have then in `__main__`

scripts/init-database.py
```diff
@@ -0,0 +1,56 @@
...
+
+def create_database():
+    param = values['database']['database_type']+'://'+values['database']['database_user_username']+':'+values['database']['database_user_password']+'@'+values['database']['database_ip']
+    try:
+        engine = sqlalchemy.create_engine(param)
+	engine.execute("CREATE DATABASE IF NOT EXISTS " + values['database']['database_name'])
+    except:
+        param = values['database']['database_type']+'://root:'+values['database']['database_user_password']+'@'+values['database']['database_ip']
+	engine = sqlalchemy.create_engine(param)
+	engine.execute("CREATE DATABASE IF NOT EXISTS " + values['database']['database_name'])
```
agentmilindu: We have to create a `USER` also, right?

agentmilindu:
For the moment we can forget the `Bassa.sql` we have to replace that with db migrations.

scripts/init-database.py
```diff
@@ -0,0 +1,52 @@
...
+    global path
+    global configs
+    path= os.path.abspath(os.path.join(os.path.join(__file__,os.pardir),os.pardir))
+    stream = open(path+"/bassa.yml", "r")
+    configs = yaml.load(stream)
+
+
+def create_database():
+    try:
+        os.environ['MYSQL_ROOT_PASSWORD']
```
agentmilindu: This should be `root_password = os.environ.get('MYSQL_ROOT_PASSWORD')`

scripts/init-database.py
```diff
@@ -0,0 +1,52 @@
...
+def retreive_values():
+    global path
+    global configs
+    path= os.path.abspath(os.path.join(os.path.join(__file__,os.pardir),os.pardir))
+    stream = open(path+"/bassa.yml", "r")
+    configs = yaml.load(stream)
+
+
+def create_database():
+    try:
```
agentmilindu: Python won't throw if the asked ENV var is not set. you will simply get an empty string. Check for empty values.
image

scripts/init-database.py
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env python
+
+import yaml
+import sqlalchemy
+import os
+import getpass
+
+configs = 'configurations'
```
agentmilindu: Set these with empty strings.

scripts/init-database.py
```diff
@@ -0,0 +1,52 @@
...
+        os.environ['MYSQL_ROOT_PASSWORD']
+    except:
+        root_password = getpass.getpass(prompt = 'Enter root password:')
+        os.environ['MYSQL_ROOT_PASSWORD'] = root_password
+    connection_url = configs['database']['database_type']+'://root:'+os.environ['MYSQL_ROOT_PASSWORD']+'@'+configs['database']['database_ip']
+    engine = sqlalchemy.create_engine(connection_url)
+    engine.execute("CREATE DATABASE IF NOT EXISTS " + configs['database']['database_name'])
+    engine.execute("CREATE USER "+configs['database']['database_user_username']+"@"+configs['database']['database_ip']+" IDENTIFIED BY " + configs['database']['database_user_password'])
+    engine.execute("GRANT INSERT, UPDATE, SELECT, DELETE ON "+configs['database']['database_name']+".* TO "+configs['database']['database_user_username']+"@"+configs['database']['database_ip'])
+    engine.execute("FLUSH PRIVILEGES")
```
agentmilindu: Since you are granting permissions using `GRANT` you really don't need to `FLUSH PRIVILEGES`. But that is okay.

scripts/init-database.py
```diff
@@ -0,0 +1,51 @@
...
+        root_password = getpass.getpass(prompt = 'Enter root password:')
+        os.environ['MYSQL_ROOT_PASSWORD'] = root_password
+    connection_url = configs['database']['database_type']+'://root:'+os.environ['MYSQL_ROOT_PASSWORD']+'@'+configs['database']['database_ip']
+    engine = sqlalchemy.create_engine(connection_url)
+    engine.execute("CREATE DATABASE IF NOT EXISTS " + configs['database']['database_name'])
+    engine.execute("CREATE USER "+configs['database']['database_user_username']+"@"+configs['database']['database_ip']+" IDENTIFIED BY " + configs['database']['database_user_password'])
+    engine.execute("GRANT INSERT, UPDATE, SELECT, DELETE ON "+configs['database']['database_name']+".* TO "+configs['database']['database_user_username']+"@"+configs['database']['database_ip'])
+
+
+def import_SQL():
```
JaDogg: Make SQL -> sql. Do not mix capitals in snake case. 😄

padamchopra: Done sir :+1:

scripts/init-database.py
```diff
@@ -0,0 +1,51 @@
...
+    fd = open(path+"/Bassa.sql", 'r')
+    sql_File = fd.read()
+    fd.close()
+    sqlCommands = sql_File.split(';')
+    for command in sqlCommands:
+        try:
+            with engine.connect() as con:
+                con.execute(command+';')
+        except Exception as exception:
+            pass;
```
vivonk: This one is making the code to not pass codecy tests, to make sure just use `logging` or `print` that uses exception like this
```
print('init-database: exception: %s ', exception)
```
```
import logging
LOG_FILENAME = 'relative/path/to/component/core/debug-server.log'
logging.basicConfig(filename=LOG_FILENAME,level=logging.DEBUG)
logging.debug('init-database: exception: %s ', exception)
```

agentmilindu:
@malithsen @rehrumesh Can we merge this PR?

kmehant:
@vivonk This is useful, shall we merge this?

vivonk:
@agentmilindu @JaDogg @kmehant can you please review this one?

scripts/init-database.py
```diff
@@ -0,0 +1,51 @@
...
+        try:
+            with engine.connect() as con:
+                con.execute(command+';')
+        except Exception as exception:
+            print('init-database: exception: %s ', exception)
+
+if __name__ == "__main__":
+    retreive_values()
+    create_database()
+    import_sql()
```
kmehant: shall we use subprocess module to insert the schema?
@vivonk @agentmilindu @JaDogg what are your thoughts? Else we can go with the present implementation.

JaDogg: Current version is fine. Probably need to execute as a whole file rather than splitting it and running it line by line.
Parsing SQL is a new problem we don't really want to solve. (See: https://github.com/andialbrecht/sqlparse) << we can use this library if we really want to split and execute line by line.

Issue #313, opened 2017-12-22
Use Gulp Webserver Proxies to map the API to `/api`


agentmilindu:
https://github.com/scorelab/Bassa/blob/b74c25d6012c525cc41fc5ed9d4a0dcca184c6b8/ui/src/app/index.js#L6

Instead `localhost:5000`, we need to have this value as `http://localhost/api`, but for this, we have to proxy the Flash API running on port 5000 to `/api`

iammosespaulr:
I can do that.

iammosespaulr:
a small problem though ... if you change BassaUrl to localhost/api then what about the POST requests from DashBoardService.js 
```return $http({
          method: 'POST',
          url: BassaUrl + '/api/download',
          data: JSON.stringify(link),
          headers: {'Content-Type': 'application/x-www-form-urlencoded'}```
 where there is an extra /api
so wouldn't the url become localhost/api/api/download and wouldn't the
 likewise happen for all others ?

PR #355, opened 2017-12-27
Fix issue "Use Gulp Webserver Proxies to map the API to `/api` #313"

iammosespaulr:


ui/src/app/index.js
```diff
@@ -3,7 +3,7 @@
 angular.module('bassa', ['ngAnimate', 'ngCookies', 'ngTouch',
   'ngSanitize', 'ui.router', 'ngMaterial', 'nvd3', 'app'])
 
-  .value('BassaUrl', 'http://localhost:5000')
+  .value('BassaUrl', 'http://localhost/api')
```
agentmilindu: I think if we change this to just `/api` this will work across all the scenarios even when the host is not `localhost`.

PR #358, opened 2017-12-28
Create a PROD Docker for Bassa API

kmehant:


components/core/Dockerfile
```diff
@@ -1,19 +1,35 @@
...
 ARG BUILD_DATE
 ARG VCS_REF
 
 WORKDIR /home/Bassa
 EXPOSE 5000
 COPY . .
+EXPOSE 3031
+
+RUN apk add --no-cache \
+    uwsgi \
```
agentmilindu: You might need `&&` on these two lines also I guess.

components/core/Dockerfile
```diff
@@ -1,19 +1,35 @@
...
+    && ./setup.sh \
+    && python setup.py develop 
 
-RUN apt-get update \
-&& ./setup.sh \
-&& python setup.py develop 
+COPY uwsgi.ini /etc/uwsgi/
 
-CMD [ "python", "Main.py" ]
+CMD [ "uwsgi", "--socket", "0.0.0.0:3031", \
```
agentmilindu: Let's stick to `5000` instead `3031`

components/core/Dockerfile
```diff
@@ -1,19 +1,35 @@
...
+               "--plugins", "python3", \
+               "--protocol", "uwsgi", \
+               "--wsgi", "main:application", \
```
agentmilindu: I'm not sure if it is `application` we have to call.

components/core/Dockerfile
```diff
@@ -1,19 +1,35 @@
...
+      "python", "Main.py"]
```
agentmilindu: Are we running `python Main.py` command again? Or are we passing them as params to uWSGI?

components/core/Dockerfile
```diff
@@ -1,19 +1,35 @@
...
+
+RUN apk add --no-cache \
+    uwsgi \
+    uwsgi-python3 \
+    && python3 -m ensurepip \
+    && rm -r /usr/lib/python*/ensurepip \
+    && pip3 install --upgrade pip setuptools \
+    && pip3 install -r requirements.txt \
+    && rm -r /root/.cache \
+    && ./setup.sh \
```
agentmilindu: Can't we move all the lines into `setup.sh` and simply just `RUN ./setup.sh`?

Issue #364, opened 2017-12-29
Creating an installation script for Bassa


amittal27:
The instructions provide an outline for how to install Bassa onto the local machine; however, it may be hard to follow if the user is not already experienced with terminal, python, or git. Thus, it would be incredibly beneficial if the instructions were more detailed so users who aren't as familiar with the instructions specified in README.md for the installation procedure.

## Expected Behavior

I'm suggesting that we create installation scripts to make the process for installing Bassa onto one's machine much quicker and easier. In this way, the user doesn't need much technical experience to use Bassa. 

## Current Behavior

My suggestion for creating an installation script will allow the user to encounter fewer obstacles and eliminate frustration when installing Bassa.

## Possible Solution

I think the process for installing Bassa should be seamless. This can be done by creating installation scripts designated for specific operating systems, such as Windows (7, 8, 10), Mac OS X, and Linux (such as Ubuntu, Fedora, and Arch). Instead of having to follow a series of instructions (which do not provide all of the necessary details to install the software), the user just needs to download and run the installation script.

I have already started working on an installation script for installing Bassa onto a machine running Ubuntu 16.04. I would be happy to submit a pull request of the script after I refine it.

## Context

As I was working to install Bassa as one of the tasks for Google Code-in 2017, I ran into may difficulties that made the process lengthy and somewhat confusing. In other words, installing Bassa took a lot longer than necessary.

## Your Environment

* Operating System and version (desktop or mobile): Ubuntu 16.04 (virtual machine)

kmehant:
Please look into this https://github.com/scorelab/Bassa/pull/374#issuecomment-498620595 
I am closing this issue for now.

PR #374, opened 2017-12-31
Installation Script for Bassa on Ubuntu 16.04

amittal27:
-- This pull request addresses issue #364 --

Bassa can be installed on Ubuntu 16.04 by running this script.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -1,24 +1,24 @@
-#/bin/bash
+#!/bin/bash
 
 echo "=== START OF FILE ==="
 
-echo "=== <Check to see if user is root> ===="
-if [ `whoami` != 'root' ]; then
+echo "==== <Check to see if user is root> ===="
+if [ $(whoami) != 'root' ]; then
 	echo "Please run this script as root or use sudo"
 	exit
 fi
-echo "=== </Check to see if user is root> ===="
+echo "==== </Check to see if user is root> ===="
 echo
 
-echo "=== <Set path to Bassa directory> ===="
+echo "==== <Set path to Bassa directory> ===="
 path=$(find ~  -name "Bassa")
 echo $path
 cd $path
-echo "=== </Set path to Bassa directory> ===="
+echo "==== </Set path to Bassa directory> ===="
 echo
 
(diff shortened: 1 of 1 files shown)


saisankargochhayat:
Please send a PR to scorelab:Bassa instead of scorelab:master

scripts/install.sh
```diff
@@ -0,0 +1,65 @@
...
+./setup.sh # run setup script
+cd components/core/ 
+python setup.py develop 
+echo "==== </BASSA Setup> ====" 
+echo 
+
+echo "==== <BASSA DB Setup> ===="
+export BASSA_DB_USERNAME=root # set the environment variables from DBCon.py 
+export BASSA_DB_PASSWORD=Bassa
+export BASSA_DB_NAME=Bassa
```
vivonk: I think this will do only for the session, not for the whole time. When you restart your terminal then it will forget the things. Also, make a Linux distribution based script, that can run on all the Linux systems, not only Ubuntu or something. Making a docker based script and then test it.  Else LGTM :)

agentmilindu:
@malithsen @rehrumesh What can we do to this PR?

scripts/install.sh
```diff
@@ -0,0 +1,65 @@
...
+echo "==== </Set path to Bassa directory> ===="
+echo
+
+echo "==== <Python change 3.5> ===="
+pythonVersion=$(python --version | grep -o "[[:digit:]]\.[[:digit:]]")
+if [ $pythonVersion = "3.5" ]; then
+   echo "Verified that you are using python 3.5"
+else
+   rm /usr/bin/python
+   ln -s /usr/bin/python3.5 /usr/bin/python
```
malithsen: This might cause problems else where in the system. Wouldn't it be better if we remember the version and use it to run the scripts

kmehant:
@amittal27 Great implementation but would have been good if it could do the job right on all the Linux distros, Darwin and Windows platforms. Thanks for the contributions. We are taking Bassa to docker environment using which we can setup Bassa literally anywhere with very few steps or even with a single command in development mode. So closing this **PR** for now. You can discuss any points in the Bassa _gitter_ channel we can help you out!

Issue #324, opened 2017-12-24
Breakdown REST.py into manageable subparts


agentmilindu:
I think we should break the `REST.py` into several parts as following,

Rename REST.py into Server.py
Create a directory as Routes/ and put an `index.py`
Create a User.py in `Routes` and add all the User related routes there.
Create a Download.py in `Routes` and add all the Download related routes there.
etc. 

And finally, import all these Route files in `Routes` directory into `index.py` and then import this Routes into new `Server.py`

vivonk:
@agentmilindu  is this issue still pending? I would like to work on this

agentmilindu:
Hmm, check the currently open PR(https://github.com/scorelab/Bassa/pull/400) by Padam. He introduces some changes, which might break the UI. Thus we have to change the UIs accordingly. if you can work on that PR and improve it to get th UIs, etc also working, that would be awesome! 

@padamchopra Can you enlightn us on this? 
@rehrumesh @malithsen

padamchopra:
@agentmilindu @vivonk , what I've done is that I've changed the code to give the response in the following way-
```
{ "status" : 200 } //success
{ "status" : 400 } //error
```

Earlier it was in the following way-
```
{ "status" : 200 } //success
{ "error" : 400 } //error
```
Due to this change in the response structure, the frontend might break. What I've done is the correct way for responses, so before merging the PR, the UI must also be worked upon

vivonk:
Bassa codebase has been refactored a lot so this issue is resolved now. @agentmilindu

PR #400, opened 2018-01-08
Break rest.py

padamchopra:
[GCI 2017]
This PR is in response to Issue #324 created by @agentmilindu

### DIFF
--- a/components/core/Routes/Download.py
+++ b/components/core/Routes/Download.py
@@ -68,10 +68,7 @@ def add_download_request():
             else:
                 newDownload = Download(data['link'], g.user.userName)
                 status = add_download(newDownload)
-                if status == "success":
-                    resp = Response(response='{"status":"'+ status + '"}', status=200)
-                else:
-                    resp = Response(response='{"error":"' + status + '"}', status=400)
+		resp = Response(response='{"status":"'+ status + '"}', status= (200 if status == "success" else 400))
         except Exception as e:
             resp = Response(response='{"error":"' + e.message + '"}', status=400)
         resp.headers['token'] = token
@@ -89,10 +86,7 @@ def remove_download_request(id):
     if token is not None:
         try:
             status = remove_download(id, g.user.userName)
-            if status == "success":
-                resp = Response(response='{"status":"'+ status + '"}', status=200)
-            else:
-                resp = Response(response='{"error":"' + status + '"}', status=400)
+            resp = Response(response='{"status":"'+ status + '"}', status= (200 if status == "success" else 400))
         except Exception as e:
             resp = Response(response='{"error":"' + e.message + '"}', status=400)
         resp.headers['token'] = token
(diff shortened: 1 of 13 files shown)


components/core/Routes/Download.py
```diff
@@ -0,0 +1,168 @@
...
+    try:
+        token = request.headers['key']
+        if str(token)!=server.config['SECRET_KEY']:
+            return "{'error':'not authorized'}", 403
+        if p is not None:
+            p.terminate()
+            p.join()
+            jsonreq = json.dumps({'jsonrpc':'2.0', 'id':'qwer', 'method':'aria2.pauseAll'})
+            jsonreq = jsonreq.encode('ascii')
+            c = urllib.request.urlopen('http://localhost:6800/jsonrpc', jsonreq)
```
rehrumesh: localhost might fail when deploying to production.

padamchopra: I did not make this :worried: 
I've just separated all the download related functions from the earlier named Rest.py into this file.

components/core/Routes/Download.py
```diff
@@ -0,0 +1,168 @@
...
+
+@server.route('/api/download/rate/<int:id>', methods=['POST'])
+def rate_download_request(id):
+    token = token_validator(request.headers['token'])
+    if token is not None:
+        data = request.get_json(force=True)
+        try:
+            status = rate_download(id, g.user.userName, data['rate'])
+            if status == "success":
+                resp = Response(response='{"status":"'+ status + '"}', status=200)
```
rehrumesh: ` resp = Response(response='{"status":"'+ status + '"}', status= (200 if status == "success" else 400))`

possible to replace the if/else block to a single line like this? Applicable for several places.

components/core/Server.py
```diff
@@ -0,0 +1,48 @@
...
+from gevent import monkey
+
+dir_path = os.path.abspath(__file__)[:-10]
+sys.path.append(dir_path+'/Routes')
+import index
+
+monkey.patch_all(ssl=False)
+
+server = Flask(__name__)
+server.config['SECRET_KEY'] = "123456789"
```
rehrumesh: SECRET is hardcoded?

components/core/Routes/Download.py
padamchopra: This change has been implemented in the latest commit @rehrumesh

components/core/Routes/User.py
```diff
@@ -171,10 +99,7 @@ def update_user_request(username):
         try:
             newUser = User(data['user_name'], data['password'], int(data['auth']), data['email'])
             status = update_user(newUser, username)
-            if status == "success":
-                resp = Response(response='{"status": "'+ status + '"}', status=200)
-            else:
-                resp = Response(response='{"error":"' + status + '"}', status=400)
+            resp = Response(response='{"status":"'+ status + '"}', status= (200 if status == "success" else 400))
```
agentmilindu: This changes the response structure, 
Previously they were
```
{ "status" : 200 }
```

```
{ "error" : 400 }
```

But your code changes it to,
```
{ "status" : 200 }
```
and
```
{ "status" : 400 }
```

However, I like it that way. But I guess that would break the frontend.

We can make this streamlined in later tasks.

rehrumesh:
@padamchopra can you rebase and push the changes please?

vivonk:
@rehrumesh I have made changes in #429 where I used blueprint so UI will not disturb.

components/core/Routes/Download.py
vivonk: @rehrumesh, please take a look [here](https://github.com/scorelab/Bassa/issues/462), I was referring the same 😊

PR #415, opened 2018-01-17
fix issue  #101

### LINKED ISSUE #101: Shell script to start Bassa servers at once
In the current project there is no script to start all the three Bassa servers at once. Therefore, writing a shell script to start Bassa servers at once will make users comfortable. I will send a PR shortly.

iammosespaulr:
Fix #101 
@malithsen @agentmilindu please review ..

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -1,0 +1,10 @@
+#!bin/bash
+cd components/core;
+gnome-terminal -e "sudo python3 Main.py";
+cd ..;
+cd ..;
+cd ui/;
+gnome-terminal -e "sudo gulp serve";
+gnome-terminal -e "sudo aria2c --enable-rpc"
+cd ..;
+echo "Bassa Running";


iammosespaulr:
@agentmilindu @malithsen Please review

start.sh
```diff
@@ -0,0 +1,10 @@
+#!bin/bash
+cd components/core;
+gnome-terminal -e "sudo python3 Main.py";
+cd ..;
+cd ..;
+cd ui/;
+gnome-terminal -e "sudo gulp serve";
```
agentmilindu: Why `gnome-terminal`?

padamchopra: @iammosespaulr I think that using this line should take care of both the servers:
```
sudo python3 components/core/Main.py & cd ui/ & sudo gulp serve;
```

kmehant:
I don't think using `gnome-terminal` to execute this would be productive and is true that not all machines have it installed. I would really appreciate if you could come with a much more simple solution to start all the servers. So closing this **PR** for now.

Issue #348, opened 2017-12-27
Check user type and show/hide admin menu


agentmilindu:
![image](https://user-images.githubusercontent.com/1391468/34372232-4d4fecbc-eaf9-11e7-8a4a-0b3df586070d.png)

Show or hide the Admin menu depending on the user type. Only Admin level users should see it.

tanmayanand44:
@agentmilindu .Can you please guide me a little to resolve this issue.

agentmilindu:
@malithsen will be you mentos for this issue :)

tanmayanand44:
Ok thanks.

tanmayanand44:
@malithsen Sir,guide me how to proceed.

malithsen:
Hi @tanmayanand44 
There are 2 user types on Bassa. 

1. Regular users
2. Admins

Admin  menu option in the side panel should only be visible to the admins. Currently it is shown to all users but only admins are allowed to view its content

agentmilindu:
@malithsen is some property set in user object to identify the user level? Where should he look for this?

malithsen:
`UserService.getAuthLevel()` will return the logged in user's auth level. 


**Auth levels** - https://github.com/scorelab/Bassa/wiki/URL-Endpoints#post
ADMIN - 0
USER - 1

malithsen:
@tanmayanand44 @agentmilindu https://github.com/scorelab/Bassa/blob/bb455d083553da38c95e9e548bb6f77354d91755/ui/src/app/index.js#L119

This is how current functionality has been implemented. Not so neat

tanmayanand44:
@malithsen Problem is that in Bassa/ui/src/app/components/services/NavService.js  all these 3 object items are being sent.We can control them then it's done.

function navService($q){
    var menuItems = [
      {
        name: 'Dashboard',
        icon: 'dashboard',
        sref: '.dashboard'
      },
      {
        name: 'Completed',
        icon: 'view_module',
        sref: '.table'
      },
      {
        name: 'Admin',
        icon: 'person',
        sref: '.admin'
      },
    ];

malithsen:
`navService.loadAllItems()` is used by `MainController`, you can filter the items there based on role 
https://github.com/scorelab/Bassa/blob/bb455d083553da38c95e9e548bb6f77354d91755/ui/src/app/controllers/MainController.js#L19-L23

tanmayanand44:
[...]

PR #424, opened 2018-02-04
Hide Admin tab for normal users

padamchopra:
### In reference to #348 with label of GoogleCodeIn
This checks the user type of the logged in user. If it is a `normal` user, it hides the `Admin` tab and if its an `admin` user, it shows the tab. 
### Screenshot of Bassa for a normal user:
screenshot from 2018-02-05 00-59-58

Another person with handle tanmayanand44 had worked on this, but his code had some bugs ("`I am experiencing a bug from my recent PR`"). My version is totally bug free in the scenarios I've tested it on.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,6 @@ node_js:
   - "6.10.0"
 python:
   - "3.6"
-
 sudo: required
 
 services:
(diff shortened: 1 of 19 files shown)


agentmilindu:
@rehrumesh @malithsen Can we merge this PR?


CC: @padamchopra

malithsen (changes requested):
This should be done on the view itself with a `ng-if`

mpiplani (approved):

PR #461, opened 2018-05-28
Merging latest to develop

agentmilindu:
Because we have done this wrong way, our `develop` is behind the `master`.


tharindupr:
Master is up to date now

components/core/Server.py
```diff
@@ -0,0 +1,46 @@
...
+from multiprocessing import Process
+from DownloadDaemon import starter
+from EMail import send_mail
+import sys
+from gevent import monkey
+
+monkey.patch_all(ssl=False)
+
+server = Flask(__name__)
+server.config['SECRET_KEY'] = "123456789"
```
rehrumesh: hardcoded secret?

rehrumesh:
Mistakenly approved. Please review the secret. Rest seems to be fine.

components/core/Server.py
vivonk: I think it's kind of private key for verifying auth tokens. It can be kept on some secret files but it would be hardcoded anyway, right? @agentmilindu @rehrumesh

agentmilindu: Yes, this secret should not be hardcoded like this in the code. It should go to the `.env` file.
But because this PR is about syncing the commits, shall we merge this and add a separate bug to fix this?

rehrumesh: Yes. Lets procced like that.

Issue #453, opened 2018-05-18
As a User, I should be able to click on the Download button and download the file


agentmilindu:
Similar to Google drive for single and multiple files to download

vivonk:
Solved in #464 , moved it from in-progress to done in GSoC 2018 Milestone

PR #464, opened 2018-05-31
Download single file from server in a single click

vivonk:
## Feature Implementation

After this commit changes, a user will able to download the files from the server to the local system.

## Description
Implemented download feature on backend and provided single click download feature on the front end

## Related Issue

#453 

## Motivation and Context

Enhancing the Bassa and implement the main feature which makes Bassa better
## How Has This Been Tested?

Tested with Safari, Mozilla, and Chrome. Tested on cross-platforms.
Run standard tests: `npm test`

image
image

## Types of changes

## Checklist:

## TODO:
- [ ] Add multiple file download feature
- [ ] Write Unit tests for changes
- [ ] Add Auth System to the file sending API


components/core/file_sender/sender.py
```diff
@@ -0,0 +1,25 @@
...
+        download_name = get_download_name_from_id(id)
+    except Exception as e:
+        logging.error(" get file API (/api/get/file) got wrong arguments, thrown an error >> " + str(e) + " <<")
+        return Response("error", status=200)
+
+    try:
+        return send_file(filename_or_fp=download_path, attachment_filename=download_name,
+                         as_attachment=True)
+    except Exception as e:
+        logging.error("File sending has a exception. We got >> "+str(e)+" << error on file sending")
```
agentmilindu: Let's not get the `str(e)` to a middle of a sentence, but let's have it as the end f the sentence.

vivonk: done 👍

components/core/file_sender/sender.py
```diff
@@ -0,0 +1,24 @@
...
+        download_name = get_download_name_from_id(id)
+    except Exception as e:
+        logging.error(" get file API (/api/get/file) got wrong arguments, thrown an error :: %s" % e)
+        return Response("error", status=200)
+
+    try:
+        return send_file(filename_or_fp=download_path, attachment_filename=download_name,
+                         as_attachment=True)
+    except Exception as e:
+        logging.error("File sending has a exception. We got :: %s on file sending" % e)
```
agentmilindu: I think you have missed this error to get to the end.

vivonk: I think it's correct? Are you saying that this error should be placed after last return?
Actually error returns some text in response so it's that return statement

agentmilindu: No, this is relevant to my previous comment about placing the error object in the log message. 

```
logging.error("File sending has a exception. When sending file, we got :: %s" % e)
```

vivonk: oh, sorry I got it wrong. So sorry, ok let me change that

Issue #454, opened 2018-05-18
The user should be authorized before downloading the files from the server


agentmilindu:
Only person having user account can download the files from server (ex. Only the person can download the files from Drive who have access to it.)

PR #465, opened 2018-06-02
Add authorization to the file download request

vivonk:
Feature Implementation

## Description

In the previous PR #464, the user was not authorized to download a file from the server. In this PR, I have added existing authorization (token based) system to the file sending API. 

## Related Issue

#454  

## How Has This Been Tested?

Tested within browsers with valid and invalid tokens and working fine ☺️ 
Run standard tests `npm test`

## Types of changes

## Checklist:

### DIFF
--- a/components/core/Server.py
+++ b/components/core/Server.py
@@ -44,7 +44,3 @@ def serve_ui1():
 
 # Import routes
 from routes import *
-from file_sender import sender
-
-# register endpoints
-server.add_url_rule('/api/get/file', 'send_file_from_path', sender.send_file_from_path, methods=['GET'])
--- a/components/core/Server.py
+++ b/components/core/Server.py
@@ -44,4 +44,7 @@ def serve_ui1():
 
 # Import routes
 from routes import *
-from file_sender import *
+from file_sender import sender
+
+# register endpoints
+server.add_url_rule('/api/get/file', 'send_file_from_path', sender.send_file_from_path, methods=['GET'])
--- a/components/core/file_sender/sender.py
+++ b/components/core/file_sender/sender.py
@@ -1,9 +1,9 @@
-from Server import *
-from flask import Response
-from Main import logging
+from flask import Response, request, send_file
+from DownloadManager import get_id_from_gid, get_download_path, get_download_name_from_id
+import logging
+from Server import token_validator
 
 
-@server.route("/api/get/file", methods=['GET'])
 def send_file_from_path():
     token = token_validator(request.args['token'])
     if token is not None:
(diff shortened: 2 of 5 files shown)


components/core/file_sender/sender.py
```diff
@@ -5,20 +5,24 @@
 
 @server.route("/api/get/file", methods=['GET'])
```
malithsen: REST of our API endpoints are in `REST.py`. How about we define the endpoint there and have the internals here?

vivonk: Ya, it seems a good way. I guess blueprint module can help us in this way [here](http://flask.pocoo.org/docs/1.0/blueprints/)

agentmilindu: I wonder if we really need Blueprints as the API is simple and have just few endpoints.

vivonk: Either, we can add a task of restructuring the APIs in a logical way. Else the APIs are already defined in a good way and structure of project is also good

vivonk: So I guess we can merge this PR? @agentmilindu

agentmilindu: @malithsen Please give your comments on this.

malithsen: Since our API is simple enough, let's just define the endpoints in REST.py and have the functionality implemented in modules. As for this PR let's move the highlighted line to REST.py and keep the internals here

vivonk: I guess this is what we can use 
http://flask.pocoo.org/docs/0.12/api/#flask.Flask.add_url_rule

vivonk: I have changed the code, please review it back @malithsen 
Also please suggest better ways to do this 😊

components/core/Server.py
```diff
@@ -44,4 +44,7 @@ def serve_ui1():
 
 # Import routes
 from routes import *
-from file_sender import *
\ No newline at end of file
+from file_sender import sender
+
+# register endpoints
+server.add_url_rule('/api/get/file', 'send_file_from_path', sender.send_file_from_path, methods=['GET'])
```
malithsen: Why do we register the endpoint here instead of REST.py?

malithsen: Oh! Where have we defined the rest of  `/api/*` endpoints in the current version?

vivonk: It's in routes 😊

malithsen: Great. In that case, let's have this in routes/Downloads.py. In the meantime let's change endpoint from `api/get/file` to `api/file` (or something suitable) as well. Usually, endpoints should consist of nouns. https://restfulapi.net/resource-naming/

vivonk: Ok, I have added new commit which have endpoint name `api/file` and endpoint is registered in Download.py

vivonk:
@agentmilindu please merge this PR 😊

Issue #456, opened 2018-05-18
As Users, we should be able to download same file/files from a single link


agentmilindu:
A single link will download file/files from server to local (similar to Google Drive file sharing system)

PR #479, opened 2018-08-05
File sharing system with multi threading

vivonk:
As in enterprise or colleges, sharing files though a link is very common thing and we don't have to keep away our user from file sharing feature.
## Description

Implemented file sharing system where user will get a zip file from the link which will contain all the necessary files that are shared.
It would look like this `https://<dns>/api/file?gid=<some alphanumeric string>&share=true`
## Related Issue
#456 

## How Has This Been Tested?
Run standard tests and cross-browser check

## Types of changes

## Checklist:

# TODO
- [ ] File click is only on list view, have to do with grid view too
- [ ] Generate link when file is actually read to get from server side


ui/src/app/completed/controllers/TableCtrl.js
```diff
@@ -49,17 +65,99 @@
...
+      if($scope.checkedFileArray.length === 0){
+        ToastService.showToast("Please checkbox some files to share");
+        return;
+      }
+
+      TableService.startCompression($scope.checkedFileArray).then(function (response) {
+        $mdDialog.show(progressDialog);
+        if(response.data['process_id'] != null) {
+          $scope.shareLink = BassaUrl;
+          $scope.shareLink += "/api/file?gid="+response.data['process_id']+"&share=true";
```
malithsen: How long is the proces_id? would someone be able to mine downloads of others by trying out random process_ids?

vivonk: process_id is the combination of gid's which means for every file cluster or single file, it's different.
I have kept the limit of 252 characters. mining gid combination would not be easy

PR #485, opened 2018-09-30
Download input field now goes to default state when a download link is added (fixes #484)

### LINKED ISSUE #484: Add download input field must return to untouch state after adding any link(UX)
Add download input field must return to untouch state after adding any link instead of validation error(red) state. 

## Expected Behavior

After adding the download link the input field must set to untouched form state.

## Current Behavior

It goes in validation error state.(RED)
red

aakashbansal:
The updated code fixes the issue!

Check the following screenshot to see :

ui-patch-1


vivonk:
What if I don't put any link and try to add a URL. I guess required was making some sense.

PR #491, opened 2018-10-01
Readme updated to explicitly specify the changes required in DBCon.py file

aakashbansal:


vivonk:
actually, if someone is using the environment variables for the DB connection then don't need to change. So specify it that if you are not using environment variables then use something like this. 
All the way Nice job :+1:

aakashbansal:
Thanks for the suggestion. I have updated the readme to take into account the environment variables configuration.

README.md
```diff
@@ -62,6 +62,17 @@ mysql -u root -p  Bassa < Bassa.sql
 
 3. Open components/core/DBCon.py and setup database username and password.
 
+The required changes in DBCon.py would look like :
+
+If the environment variables are being used, modify as following :
+```
+_db=MySQLdb.connect("db", os.environ.get('YOUR_DB_USERNAME'), os.environ.get('YOUR_DB_PASSWORD'), os.environ.get('Bassa'))
```
agentmilindu: Shall we change `os.environ.get('Bassa')` into `os.environ.get('YOUR_DB_NAME')`?

kmehant:
@aakashbansal closing this as it is fixed now 😄

PR #500, opened 2018-10-03
Duplicate links removed when adding/deleting downloads (fixes #489)

### LINKED ISSUE #489: Temporary duplicated file links when adding or removing downloads
Whenever the user adds new download link or deletes the existing one, the list gets updated. But while the list is being updated, there is some duplicacy of the file links.

Please refer the screenshot :

temp-repeated-files

aakashbansal:
Now when the user adds or deletes a download link, he won't see momentary duplicates. The "track by $index" in the ng-repeat makes sure of that. Also one modification done in code is when the download link gets deleted, instead of pinging the server to get refreshed downloads list, we just delete it from the local array, provided the server update is successful. We can't do this for new files addition(updating the array locally after new download addition) because there are some properties in the array that are set by the server. So server pinging is required.
temp-repeated-files


aakashbansal:
This PR fixes #489 . @vivonk Please have a look.

VibhorCodecianGupta:
Good stuff m8!

ui/src/app/dashboard/controllers/DashCtrl.js
```diff
@@ -69,14 +69,14 @@
...
 
-    $scope.removeLink = function (id) {
+    $scope.removeLink = function (index, id) {
       // Shows window confirmation before deleting download.
       var deleteDownload = $window.confirm('Confirm download task deletion !');
 
       if(deleteDownload){
         DashService.removeDownload(id).then(function(response){
           ToastService.showToast("Download removed");
-          getActiveDownloads();},
```
vivonk: Before removing this, just check the response. If the response is ok, means HTTP 200 then remove index else you have to show that something is wrong (A `ToastService.showToast()`).
Check the function that runs on delete request.
https://github.com/scorelab/Bassa/blob/develop/components/core/routes/Download.py#L69
Btw, Nice job man

aakashbansal: Hey thanks for pointing that out. Applied the necessary changes.

PR #504, opened 2018-10-04
Reminder to get 'Go' from maintainers before a PR

agentmilindu:
We see a lot of first-time contributors creating issues and developing them without properly discussing with the project maintainers. 

This is sometimes problematic, thus reminding the issue creators to first have a good discussion with maintainers is a good way to avoid this.


vivonk:
Strongly agree with you @agentmilindu

PR #521, opened 2018-10-19
Feature/automate db scripts in docker compose

gnomezgrave:
Automate the DB initialization when using docker-compose

## Description
One of the best ways to isolate our code from the OS level dependencies is to dockerize the code. Even though there is a `docker-compose.yml` and it creates the mysql server, it does not initialize the database and the tables needed to run Bassa core.

This PR will add a shell script for the `docker-entrypoint-initdb.d` and it will run the `Bassa.sql` on the docker db.

## Related Issue
No issue was open for this task since this is an improvement.

## Motivation and Context
This will reduce a manual step required to get the system deployed.

## How Has This Been Tested?
Tested on the local machine in several runs. This script runs only when the `mysql` docker image is created.

N/A

## Types of changes

## Checklist:


.docker/volumes/mysql/create-schema.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
+SCHEMA_DIR=/tmp/schema
+
+mysql_exec=( mysql -uroot -proot )
```
agentmilindu: Are we using `root` as the MySQL root password here?
I guess root password can be changed by setting a value to `MYSQL_ROOT_PASSWORD` ENV variable.
Shall we use `${MYSQL_ROOT_PASSWORD}` as the root password?

gnomezgrave: Damn! Nice catch!

agentmilindu:
@vivonk please approve

PR #576, opened 2018-10-28
Update setup.py to be compatible with pip10

Benjamin-Mueggenburg:
Setup.py is now compatible with pip10

## Description
Setup.py was not compatible with pip version 10. 

## Motivation and Context
Could not run setup.py - was giving import error

### DIFF
--- a/components/core/setup.py
+++ b/components/core/setup.py
@@ -3,7 +3,10 @@ import shutil
 import inspect
 import platform
 from setuptools import setup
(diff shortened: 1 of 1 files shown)


malithsen:
This will tentatively fix the problem. But we shouldn't be relying on internal API calls of pip. Proper way to resolve this should be using `packages` in `setup` https://python-packaging.readthedocs.io/en/latest/dependencies.html

PR #683, opened 2018-12-16
Single array to filter queued downloads

### LINKED ISSUE #503: Conserving memory while making API calls from the frontend
## Current Behavior
While making an API call to the backend which fetches all the downloads (queued or downloaded), the code uses `filter` function to filter the queued downloads and then maps them. This process takes up a lot of extra space as both `filter` and `map` functions create new arrays, hence allocating extra memory twice for the same task. Since this API is being called every time a user adds a link to be downloaded, it becomes pretty memory inefficient.
Code snippet: https://github.com/scorelab/Bassa/blob/1294190d79cc00f98e0579455564634ac858b015/ui/src/app/dashboard/controllers/DashCtrl.js#L63

## Expected Behavior
Minimal and judicious use of memory is the goal.

## Possible Solution
Using the `reduce` function on the response array, instead of `filter` and `map`. The reduce function, written accordingly, can easily carry out the task of filter and map in the same iteration, and it returns the very same array without creating new ones, and hence saves memory.

Resources:
https://codeburst.io/reduce-filter-and-map-without-reduce-filter-and-map-787c21671c95
https://medium.freecodecamp.org/reduce-f47a7da511a9
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce

VibhorCodecianGupta:
## Description
Used `reduce` function instead of mapping and filtering to avoid the creation of two distinct arrays on every download queue.

## Related Issue
fixes #503 

## Motivation and Context
A lot of memory will unnecessarily be created on adding every download since two arrays are used for filtering and mapping. The objective was to combine the tasks to take place in a single iteration.

## Types of changes

## Checklist:


ui/src/app/dashboard/controllers/DashCtrl.js
```diff
@@ -52,18 +52,21 @@
...
     $scope.addLink();
   }
     var getActiveDownloads = function() {
       DashService.getDownloads().then(function (response) {
         var data = response.data;
-        $scope.downloads = _.filter(data, function(d) {return d.status==0});
-        $scope.downloads = _.map($scope.downloads, function(element) {
-             return _.extend({}, element, {progress: 0});
-        });
+        $scope.downloads = _.reduce(data, function(queue, d) {
```
vivonk: Agree :+1:

vivonk: It doesn't break any functionality? just test it out once.

VibhorCodecianGupta: Tested, works fine! :D

PR #685, opened 2018-12-19
Add Decline Request Button

### LINKED ISSUE #496: Adding reject option for rejecting signup of any user if needed.
Add an option for ejecting the signup of any specific user if needed

## Expected Behavior
There is no such option in the project currently, it simply accepts all the signup requests. So, we could have a 'reject' feature too in order to reject the sign of a user when needed.

## Current Behavior
No such feature in the project right now.

anuragparida:
Added a Decline Request Button to the Requests Section in the Admin Tab, since right now there is no way of declining a request, only for accepting.

Decline

fixes #496


anuragparida:
@agentmilindu @charithccmc Please review and shed some light on what might be going wrong? :slightly_smiling_face:

agentmilindu:
I think @malithsen and @vivonk would be the best people to help you on this

anuragparida:
Sure, thanks! :slightly_smiling_face: @malithsen @vivonk PS- New additions to AdminCtrl.js and AdminService.js are at line 58 and line 35 respectively, my linting software produced spaces everywhere xD. I will fix that in the final styling commit :slightly_smiling_face:

vivonk:
I mean there was already a cancel button where you have added new functionality. A button to cancel sign-up request. Wasn't that working?

anuragparida:
I'm sorry Sir, but I don't think there was any old functionality for canceling the request. I merged the develop branch right before starting to code, so I don't think I have missed it. Could you recheck? There is no button for canceling in the [Current Develop Branch](https://github.com/scorelab/Bassa/blob/develop/ui/src/app/views/admin.html) Admin View either :neutral_face:

anuragparida:
Check this out https://github.com/scorelab/Bassa/issues/496

vivonk:
Ok my fault, I got mismatched with download request. Sorry.
Yup, your works LGTM :smile:

malithsen:
I'm not sure deleting the requests from the db is a good option because it is an irreversible action. What if the admin mistakenly deletes the requests or for some use reason he/she later wants to revoke the signup rejection? This way we can also avoid someone spamming the system with signup requests.

How about we simply change the status of the request. Then have another tab to see rejected requests with the option to revoke?

Thoughts? @agentmilindu @vivonk

vivonk:
Agree on the @malithsen point

anuragparida:
@malithsen I agree, but wouldn't that need a change to the entire database? One alternative I can suggest is to confirm the action through an alert to ensure that there is no misclick. If the user genuinely deletes it and then regrets it later, the client can re-register

anuragparida:
@malithsen This would require least amount of messing around with other code for achieving that effect I think. Also, we would have to develop a separate section for "requests that were declines but can be restored" or basically a request recycle bin view, which is redundant in my opinion

vivonk:
@anurag deleting data from database if it's important is always a bad practice. Setting a flag makes more sense so that we don't loose any data. Altering table is not a big deal as we are currently developing Bassa not in release mode.

malithsen:
@anuragparida  DB already has a blocked column. https://github.com/scorelab/Bassa/blob/master/Bassa.sql#L63
Take a look at `/api/user/blocked` https://github.com/scorelab/Bassa/wiki/URL-Endpoints

anuragparida:
@malithsen @vivonk Oh I didn't see the blocked function. Should I change the code to refer to the blocked function directly?

anuragparida:
@vivonk I understand now, thank you :smile:

malithsen:
@anuragparida Yeah simply changing the blocked status to 1 instead of deleting the user should work 
@vivonk

anuragparida:
Okay sure, I'll implement that. But one doubt - Does the ng-repeat get the data only that is not blocked AND not approved? The first condition will be important to delete it from the table view

vivonk:
Not sure what you asking. Check the relavant backend endpoint and change query in that to not get irrelevant data.

anuragparida:
Okay, cool, will do that, Thankyou @malithsen @vivonk :smile:

agentmilindu:
@anuragparida Let's resolve conflicts of this one too

ui/src/app/admin/services/AdminService.js
```diff
@@ -1,53 +1,62 @@
...
-        headers: {'Content-Type': 'application/x-www-form-urlencoded'}
+        headers: { 'Content-Type': 'application/x-www-form-urlencoded' }
+      });
+    }
+
+    var decline = function (username) {
+      return $http({
+        method: 'POST',
+        url: BassaUrl + '/api/user/block/' + username,
+        headers: { 'Content-Type': 'application/x-www-form-urlencoded' }
```
vivonk, line 43: **Bug: Wrong endpoint URL**

The `decline` function calls `/api/user/block/` instead of `/api/user/decline/`. This means clicking Decline will actually hit the block endpoint (or 404) rather than the new decline route. Change to `BassaUrl + "/api/user/decline/" + username`.

vivonk (commented):
**PR Review: Add Decline Request Button**

Overall the feature direction is right, but there are a few bugs and issues that should be fixed before merging. Key concerns: a wrong URL in the new `decline` service function, a `splice` bug that incorrectly removes items from the list, a potential missing auth check on the backend endpoint, and blocked users silently disappearing from the queue.

components/core/UserManager.py
```diff
@@ -178,7 +178,7 @@ def get_signup_requests():
     db = threadpool.connect()
     if db is not None:
         cursor =  db.cursor(MySQLdb.cursors.DictCursor)
-        sql = "SELECT user_name, email FROM user WHERE approved=0;"
+        sql = "SELECT user_name, email FROM user WHERE approved=0 AND blocked=0;"
```
vivonk, line 181: **Logic: Declined users silently disappear from the admin queue**

Adding `AND blocked=0` hides any `blocked=1` user from the signup list, but they are never deleted, notified, or surfaced elsewhere. A declined user ends up in invisible limbo with no audit trail. Consider adding a `declined` status column, sending a rejection email, or logging the action for admin visibility.

ui/src/app/admin/services/AdminService.js
vivonk, line 43: **Security: Username concatenated into URL without encoding**

Concatenating `username` directly into the URL path without `encodeURIComponent()` can break requests or allow path manipulation if a username contains `/`, `?`, or `#`. Use `encodeURIComponent(username)` here — same fix applies to the `approve` function above.

ui/src/app/admin/controllers/AdminCtrl.js
```diff
@@ -1,67 +1,77 @@
...
+    $scope.decline = function (user) {
+      AdminService.decline(user.user_name).then(function (response) {
+        ToastService.showToast('Declined', user.user_name);
+        $scope.signup_requests.splice(user, 1);
+        getRequests();
+      }, function (error) {
+        ToastService.showToast('Oops! Something went wrong');
+      });
+    }
+
```
vivonk, line 67: **Bug: `splice` called with object instead of index**

`splice(user, 1)` passes a user object as the first argument; `Array.splice` expects a numeric index. JS coerces the object to `0`, so the *first* item always gets removed instead of the declined user. Fix: `$scope.signup_requests.splice($scope.signup_requests.indexOf(user), 1)` or just rely on `getRequests()` to refresh.

ui/src/app/admin/services/AdminService.js
```diff
@@ -1,53 +1,62 @@
...
         url: BassaUrl + '/api/user/approve/' + username,
-        headers: {'Content-Type': 'application/x-www-form-urlencoded'}
+        headers: { 'Content-Type': 'application/x-www-form-urlencoded' }
+      });
+    }
+
+    var decline = function (username) {
+      return $http({
+        method: 'POST',
+        url: BassaUrl + '/api/user/block/' + username,
```
vivonk, line 42: **Bug: Decline is calling the wrong backend endpoint**

The URL is `/api/user/block/` but the route registered in `Server.py` is `/api/user/decline/<username>`. Clicking Decline will hit the wrong endpoint. Change to `BassaUrl + "/api/user/decline/" + username`.

PR #687, opened 2018-12-21
Positioned Buttons in AdminView

anuragparida:
Set the Approve Button in Admin Table to be centred vertically because it was looking odd before.
Screenshots attached-
PS- Screenshots have additional feature (different PR) but code change has only centering change.


anuragparida:
@malithsen @vivonk Please review :smile:

vivonk:
@anuragparida Make an issue for it as well so that later we can have a reference to discuss anything regarding design changes of Admin tab.

vivonk:
@anuragparida Also, people don't use to dig into the code for the first time so for better understanding add a simple code snippet in the description so that people get to know what exactly you are trying to do and then suggest changes. So let's try this way for #687 and later to #687

PR #688, opened 2018-12-21
Remove only approved user from AdminView

### LINKED ISSUE #686: Approve User Button always removes the first element in table
When I click on any approve button in the Requests Section of the Admin Tab, it ALWAYS removed ONLY the first element from the table.

I clicked-

It removed first element-

However, this is only in the UI, because when I reload the page/switch to a different tab and switch back to Admin Tab, the correct view is shown.

anuragparida:
Earlier the approve button used to remove only the first element in the Signup Requests table. Now it removes the correct element, only the one which is clicked.

Fixes #686

### DIFF
--- a/ui/src/app/admin/controllers/AdminCtrl.js
+++ b/ui/src/app/admin/controllers/AdminCtrl.js
@@ -1,8 +1,8 @@
-(function(){
+(function () {
   'use strict';
   angular
     .module('app')
-    .controller('AdminCtrl', [ '$scope', 'ToastService', 'AdminService', 'UtilityService', AdminCtrl]);
+    .controller('AdminCtrl', ['$scope', 'ToastService', 'AdminService', 'UtilityService', AdminCtrl]);
 
   function AdminCtrl($scope, ToastService, AdminService, UtilityService) {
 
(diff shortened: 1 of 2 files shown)


anuragparida:
@malithsen @vivonk

agentmilindu:
@anuragparida Let's resolve conflicts

ui/src/app/admin/controllers/AdminCtrl.js
```diff
@@ -1,67 +1,67 @@
...
         ToastService.showToast('Paused all downloads');
-      }, function(error){
+      }, function (error) {
         ToastService.showToast('Oops! Something went wrong');
       });
     };
 
-    $scope.approve = function(user) {
+    $scope.approve = function (user) {
       AdminService.approve(user.user_name).then(function (response) {
```
vivonk: @malithsen are we only intended to approve a request, not to deny because in Admin service and Backend we haven't written any endpoint to deny a signup request. So should we create an endpoint or let's leave it or now?

ui/src/app/admin/controllers/AdminCtrl.js
```diff
@@ -1,67 +1,67 @@
...
         ToastService.showToast('Approved', user.user_name);
-        $scope.signup_requests.splice(user, 1);
+        $scope.signup_requests.splice($scope.signup_requests.indexOf(user), 1);
```
vivonk: Good catch @anuragparida

ui/src/app/admin/controllers/AdminCtrl.js
anuragparida: @vivonk I did create a denial system using the /block endpoint. It's in another PR

anuragparida:
@agentmilindu I'm on it, a little juiced up in exams right now, will get back to this ASAP :smile:

vivonk:
Any update on it @anuragparida

Issue #245, opened 2017-12-10
`sudo`  in `setup.sh` making problems in Docker


agentmilindu:
Docker runs the setup.sh file as root and having problems with `sudo` commands as `sudo` is not available in `root` user. This makes the setup.sh not usable in docker image creation. What can we do?

malithsen:
We can change the setup.sh to check if it's running as root and change the commands accordingly

agentmilindu:
@malithsen What you think of the apporach used in https://github.com/scorelab/Bassa/pull/690 ?

malithsen:
@agentmilindu Made few comments. Let's continue the discussion there

kmehant:
This is fixed so closing the PR 😄

PR #689, opened 2018-12-22
Made the script Docker-compatible + Prevented version failures

rohancl:
### Made setup.sh compatible for Docker and bash users:
Docker always runs as root, Earlier the "sudo" commands were making trouble for Docker but not for bash users. Docker can't access the ```sudo``` command as its $USER is already set as root. In this commit, I've removed all the ```sudo``` (s) in the shell script.
 <br>
### OK, but wouldn't this mess up the installation for bash users since they require some permission to install packages?</b>
<br>
To take care of that, I added an If-else condition for detecting the $USER, If that's found as a "non-root" user, An echo of instructions to proceed with root user is given out in the terminal, When bash users switch to SuperUser (su) and execute this script, the installation takes place flawlessly.
<br><br>
<b>EXAMPLE: WHEN ACCESSING SCRIPT AS "NON ROOT USER"</b>
<br><br>

proo

<br>
<b>EXAMPLE: WHEN ACCESSING SCRIPT AS "ROOT USER"</b>
<br><br>

proof

<b>NOTE: This implementation forces bash users to use SU, that's because linux bash users can compromise, but as for docker users, They can only install packages as root and using </b>```sudo``` <b>in the script for docker would be making trouble. So I implemented a way which could support installation for both, bash and Docker users. For the current implementation, The </b>```Please run as root to proceed with installations```<b> won't ever come in Docker because it only runs as a root user</b>
### Prevented version failures
It's important to first update package list, then update the versions of all pre-installed packages in the system. This script installs a few popular things like npm,pip,nodejs etc.. 
Many people like me are already having them. In that case, APT just ignore the installation if the presence of that specific package is detected, and this results in the operations of older versions of packages, which to an extent may cause depreciation warnings or failures. If we update and upgrade the versions of the packages, The build gets cleaner and safer. I've added two shell commands before the installation:
<b>1.</b> ```apt-get update```: To update package list, this is a good practise to update the list before installing any packages.
<b>2.</b> ```apt-get upgrade```: To get all the pre-installed packages on the latest version.

## Related Issue

#245 
## Motivation and Context

This change is required to make this installation script compatible for Docker users
## How Has This Been Tested?

This has been tested by executing the script in the terminal with 2 different users:
1. root users 
2. non-root users

Uploaded Above ^
## Types of changes

## Checklist:

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -41,6 +41,6 @@ then
 	npm install -g gulp
 	echo "gulp installed"
 else
-	echo "Please run as root to proceed with installations... \n1. Type in: su\n2. Enter your password\n3. Execute: sh setup.sh"
+	printf "Please run as root to proceed with installations... \n1. Type in: su\n2. Enter your password\n3. Execute: sh setup.sh\n"
 	exit
 fi
--- a/setup.sh
+++ b/setup.sh
@@ -1,34 +1,46 @@
 #!/bin/bash
 
-echo "Installing pip"
-sudo apt-get install -y python3-pip
-echo "pip installed"
+if [ $USER = "root" ]
+then 
+	echo "Starting Installation..."
 
-echo "Installing setuptools"
-sudo apt-get install -y python3-setuptools
-echo "pip installed"
+	echo "Udating package list and Pre-installed package versions"
+	apt-get update && apt-get upgrade
+	echo "All packages now updated and ready to be used"
 
-echo "Installing Aria2"
-sudo apt-get install -y aria2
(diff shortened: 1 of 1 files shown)


rohancl:
@malithsen @agentmilindu Please review ^.^

rohancl:
@malithsen I don't get why my Travis CI test is failing. I'm. Trying to figure it out. Can you please tell me what output does the terminal in docker environment give when you type in ```uname -a``` in it ... I am expecting the terminal to return ```root```
Thank you ^.^

PR #690, opened 2018-12-22
Made the script Docker-compatible + Prevented version failures

rohancl:
### Made setup.sh compatible for Docker and bash users:
Docker always runs as root, Earlier the "sudo" commands were making trouble for Docker but not for bash users. Docker can't access the ```sudo``` command as its $USER is already set as root. In this commit, I've removed all the ```sudo``` (s) in the shell script.
 <br>
### OK, but wouldn't this mess up the installation for bash users since they require some permission to install packages?</b>
<br>
To take care of that, I added an If-else condition for detecting the $USER, If that's found as a "non-root" user, An echo of instructions to proceed with root user is given out in the terminal, When bash users switch to SuperUser (su) and execute this script, the installation takes place flawlessly.
<br><br>
<b>EXAMPLE: WHEN ACCESSING SCRIPT AS "NON ROOT USER"</b>
<br><br>

proo

<br>
<b>EXAMPLE: WHEN ACCESSING SCRIPT AS "ROOT USER"</b>
<br><br>

proof

<b>NOTE: This implementation forces bash users to use SU, that's because linux bash users can compromise, but as for docker users, They can only install packages as root and using </b>```sudo``` <b>in the script for docker would be making trouble. So I implemented a way which could support installation for both, bash and Docker users. For the current implementation, The </b>```Please run as root to proceed with installations```<b> won't ever come in Docker because it only runs as a root user</b>
### Prevented version failures
It's important to first update package list, then update the versions of all pre-installed packages in the system. This script installs a few popular things like npm,pip,nodejs etc.. 
Many people like me are already having them. In that case, APT just ignore the installation if the presence of that specific package is detected, and this results in the operations of older versions of packages, which to an extent may cause depreciation warnings or failures. If we update and upgrade the versions of the packages, The build gets cleaner and safer. I've added two shell commands before the installation:
<b>1.</b> ```apt-get update```: To update package list, this is a good practise to update the list before installing any packages.
<b>2.</b> ```apt-get upgrade```: To get all the pre-installed packages on the latest version.

## Related Issue

#245 
## Motivation and Context

This change is required to make this installation script compatible for Docker users
## How Has This Been Tested?

This has been tested by executing the script in the terminal with 2 different users:
1. root users 
2. non-root users

Uploaded Above ^
## Types of changes

## Checklist:


setup.sh
```diff
@@ -1,49 +1,46 @@
...
-while IFS= read line
-do
-	echo "$line"
-	main_directory=$line
-done <"$main_file"
-rm -rf main_directory.txt
-echo "export PATH=\"$main_directory:$""PATH\"" | sudo tee -a ~/.bashrc > /dev/null
-source ~/.bashrc
\ No newline at end of file
+if [ $USER = "root" ]
```
vivonk: Username can be anything which has root privileges

malithsen:
Good work @rohancl
I am a bit worried about the complications this may cause for logging. Bassa made to be working in a corporate environment. When you switch to root with su, Bassa's logging will always see the running user as root. This will not happen when we use sudo.

This is not a major concern because I don't think we are/will log a lot of user level stuff. Unless there are better suggestions, let's go with this PR :)

malithsen:
How about using
```
if ! [ $(id -u) = 0 ]; then
   echo "Not root. Run normal commands"
else
   echo "root. Run docker commands"
fi
```

`id -u` is only zero when running under an account with root privileges. And it works fine with sudo eliminating the need to use `su`. I haven't tested this extensively though. @rohancl wdyt?

setup.sh
rohancl: Right, What could we do for that?

rohancl:
Wait, I think I should try using your code in my way. 
I'll edit and make a commit in few hours ...

rohancl:
@agentmilindu @malithsen @vivonk Check the latest commit. I think this way of detection provided by @malithsen is more efficient, Earlier I used a static approach of detecting the root and non-root users, this method is dynamic.

rohancl:
Thanks @vivonk ^.^

setup.sh
```diff
@@ -1,49 +1,44 @@
...
-	echo "$line"
-	main_directory=$line
-done <"$main_file"
-rm -rf main_directory.txt
-echo "export PATH=\"$main_directory:$""PATH\"" | sudo tee -a ~/.bashrc > /dev/null
-source ~/.bashrc
\ No newline at end of file
+#!/bin/sh
+if ! [ $(id -u) = 0 ]; then
+   	printf "Please run as root to proceed with installations... \n1. Type in: su\n2. Enter your password\n3. Execute: sh setup.sh\n"
```
gnomezgrave: Any reason for using `printf` instead of `echo` ?

rohancl: Codacy code quality requirement
There is no output difference btw on using echo or printf ^.^

PR #705, opened 2019-03-07
Updated Org Name in LICENSE File

vivonk:
Added necessary information like project name, organization name in the LICENSE file


vivonk:
Merging as it's not a codebase change and necessary to do before an official release.

PR #776, opened 2019-05-19
Updated dockerfile for Bassa API server with necessary changes and optimization

### LINKED ISSUE #775: Updating docker file for Bassa API server
**Is your feature request related to a problem? Please describe.**
Presently docker file for Bassa API server is failing to build due to missing of some dependency installation steps. Dockerfile Layers has to rearranged and multi-stage builds should be changed to a little extent in order to optimize the present docker file.

**Describe the solution you'd like**
* Add the necessary dependency installation steps for Pylint.
* Rearrange the existing layers for better future docker builds.
* Decouple Aria2c RPC server from Bassa API server container for better scaling in the future.
* Optimize the present multi-stage builds.

kmehant:
## Description
* Added the necessary dependency installation steps for Pylint.
* Rearranged the existing layers for better future docker builds.
* Decoupled Aria2c RPC server from Bassa API server container for better scaling in the future.
* Optimized the present multi-stage builds.
* Used a single Base image.

## Related Issue
Fixes #775 
Presently docker file for Bassa API server is failing to build due to missing of some dependency installation steps. Dockerfile Layers has to be rearranged and multi-stage builds should be modified in order to optimize the present docker file.

## How Has This Been Tested?
This has been tested locally using Docker.

## Result
A working and optimized docker file for Bassa API server.
* `image size` ~ 450mb
* Further, the docker image can be squashed before uploading it to the official SCoRe Lab image registry.
# Types of changes

## Checklist:

### DIFF
--- a/components/core/Dockerfile
+++ b/components/core/Dockerfile
@@ -26,7 +26,7 @@ RUN python3 setup.py develop
 # Container to lint python code
 FROM base as linter 
 
-# Setting current work directory to /home/Bassa
+# Setting current work directory to /Bassa
 WORKDIR /bassa
 
 # Installing pylint and linting the python code
@@ -45,7 +45,7 @@ ARG VCS_REF
 # Volume for accessing the downloaded files
 VOLUME [ "/downloads" ]
 
-# Setting current work directory to /home/Bassa
+# Setting current work directory to /Bassa
 WORKDIR /bassa
 
 # Exposing ports where the server listens 
--- a/components/core/Dockerfile
+++ b/components/core/Dockerfile
@@ -1,32 +1,67 @@
-FROM python:3-alpine as builder
+# Container to lint python code
+FROM python:3.7-slim as linter 
 
-WORKDIR /home/Bassa
+# Setting current work directory to /home/Bassa
+WORKDIR /linter
+
+# Copying API component
 COPY . .
 
-RUN ls && pip install pylint \
-&& pylint -E .
+# Installing dependencies
+RUN apt-get update && \ 
+    apt-get install -y --no-install-recommends \ 
+    python3-pip \ 
+    build-essential \
+    default-libmysqlclient-dev \ 
+    python3-dev \
+    python3-setuptools && \ 
+    rm -rf /var/lib/apt/lists/* && \
+    apt-get clean
 
 
-FROM python:3.7
+# Installing pylint and linting the python code
+RUN pip install pylint && pylint -E . 
+
+# -----------------------------
+
+# Container for running the Bassa API server
+FROM python:3.7-slim as production
 
 MAINTAINER SCoRe Lab Community <commuity@scorelab.org>
 
 ARG BUILD_DATE
 ARG VCS_REF
 
-WORKDIR /home/Bassa
-EXPOSE 5000
+# Volume for accessing the downloaded files
+VOLUME [ "/downloads" ]
+
+# Setting current work directory to /home/Bassa
+WORKDIR /API
+
+# Copying API component
 COPY . .
 
+# Installing dependencies
+RUN apt-get update && \ 
+    apt-get install -y --no-install-recommends \ 
+    python3-pip \ 
+    build-essential \
+    default-libmysqlclient-dev \ 
+    python3-dev \
+    python3-setuptools && \ 
+    rm -rf /var/lib/apt/lists/* && \
+    apt-get clean
 
-RUN apt-get update \
-&& ./setup.sh \
-# && python3 setup.py test \ disabing temporary
-&& python3 setup.py develop
+# Installing the python modules
+RUN python3 setup.py develop
 
-CMD [ "./run.sh"]
+# Exposing ports where the server listens 
+EXPOSE 5000
 
+# Start the Bassa API server
+CMD [ "python3", "Main.py" ]
 
+# Adding data as docker labels
 LABEL multi.org.label-schema.name="Bassa" \
       multi.org.label-schema.description="Bassa provides Automated Download Queue to make the best use of Internet bandwidth" \
       multi.org.label-schema.url="https://github.com/scorelab/Bassa/wiki" \
@@ -36,4 +71,3 @@ LABEL multi.org.label-schema.name="Bassa" \
       multi.org.label-schema.vendor="Sustainable Computing Research Group" \
       multi.org.label-schema.version="" \
       multi.org.label-schema.schema-version="1.0"
-      
--- a/components/core/Dockerfile
+++ b/components/core/Dockerfile
@@ -1,42 +1,8 @@
-# Container to lint python code
-FROM python:3.7-slim as linter 
+# Container as boiler plate
+FROM python:3.7-slim as base 
 
-# Setting current work directory to /home/Bassa
-WORKDIR /linter
-
-# Copying API component
-COPY . .
-
-# Installing dependencies
-RUN apt-get update && \ 
-    apt-get install -y --no-install-recommends \ 
-    python3-pip \ 
-    build-essential \
-    default-libmysqlclient-dev \ 
-    python3-dev \
-    python3-setuptools && \ 
-    rm -rf /var/lib/apt/lists/* && \
-    apt-get clean
-
-
-# Installing pylint and linting the python code
-RUN pip install pylint && pylint -E . 
-
-# -----------------------------
-
-# Container for running the Bassa API server
-FROM python:3.7-slim as production
-
-MAINTAINER SCoRe Lab Community <commuity@scorelab.org>
-
-ARG BUILD_DATE
-ARG VCS_REF
-
-# Volume for accessing the downloaded files
-VOLUME [ "/downloads" ]
-
-# Setting current work directory to /home/Bassa
-WORKDIR /API
+# Setting current work directory to /bassa
+WORKDIR /bassa
 
 # Copying API component
 COPY . .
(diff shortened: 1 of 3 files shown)


agentmilindu:
If we remove `aria2` from Bassa docker, we have to create a separate docker for that too.

kmehant:
> If we remove `aria2` from Bassa docker, we have to create a separate docker for that too.

Sure @agentmilindu !! It's part of my GSoC work where I have to properly decouple Bassa API server and Aria2c in to separate containers.

agentmilindu:
@kmehant Yeah, agree, better to divide them into two

demon-36:
@kmehant don't you think we should optimize Dockerfile to reduce Image size using python-alpine ?
@vivonk @agentmilindu what do you say ?

kmehant:
> @kmehant don't you think we should optimize Dockerfile to reduce Image size using python-alpine?
> @vivonk @agentmilindu what do you say?

@demon-36 
I would like to mention that docker image size is not just a deciding factor for optimization. In fact, we should not blindly fall into the false idea that alpine always an optimization over normal fat Debian images without proper testing. For optimization, we have to consider what you are actually containerizing. Bassa API server is written in python as a flask application. After some research (Reddit and others) and testing which I have performed, I found out alpine is not a good choice for Bassa API server docker file.
1. Alpine version deteriorates the python application performance, though the image is lightweight.
2. There exist major differences between alpine and other Linux distros(Musl and Glibc) which might cause problems.
3. The smaller default stack size for threads causes python crashes.

*References:*
* [Python crash](https://bugs.python.org/issue32307)
* [Alpine performance loses over debian- docker image](https://superuser.com/questions/1219609/why-is-the-alpine-docker-image-over-50-slower-than-the-ubuntu-image)
* [Redit](https://www.reddit.com/r/docker/comments/8s8n3o/heres_a_bunch_of_docker_best_practices_that_i/e0xsxjd/)
@vivonk @agentmilindu I need your opinions about this.

vivonk:
Correct @kmehant . Going for alpine will require us to do some testing first, make an analysis from prospective of stable run and compatibility. Firstly, check that the components which we are using in Bassa as available on alpine version and have stable release with/out minor issues. 
Your thread point is absolutely correct @kmehant 
@demon-36 @agentmilindu what would you say?

kmehant:
@agentmilindu @vivonk @charithccmc please review this!
@demon-36 Please look into this, if it looks good then we can go with this.

components/core/Dockerfile
```diff
@@ -1,32 +1,62 @@
...
+
+# Copying API component
 COPY . .
 
-RUN ls && pip install pylint \
-&& pylint -E .
+# Installing dependencies
+RUN apt-get update && \ 
+    apt-get install -y --no-install-recommends python3-pip build-essential && \
+    apt-get install -y default-libmysqlclient-dev python3-dev && \
```
agentmilindu: Do we need another `apt-get install`? Can't we put these stuff also into the first `apt-get install`?

components/core/Dockerfile
```diff
@@ -1,32 +1,62 @@
...
+WORKDIR /home/Bassa
+
+# Installing pylint and linting the python code
+RUN pip install pylint && pylint -E . 
 
+# -----------------------------
 
-FROM python:3.7
+# Container for running the Bassa API server
+FROM scratch as production
```
agentmilindu: Can we run Puthon scripts on Scratch images? Did I miss something?

components/core/Dockerfile
kmehant: Sure we can do that 👍.  Will install them all under one single command

> Do we need another `apt-get install`? Can't we put these stuff also into the first `apt-get install`?

components/core/Dockerfile
kmehant: Yes @agentmilindu  please look into line 45 we are using multi-stage builds

agentmilindu: Yeah, but the production image is scratch, right? Which means the final image will be a scratch based image

kmehant: Not exactly @agentmilindu we can have python:slim as base image too but will be no use as we are copying the complete build (line 45) in to production container. So it can have almost nothing as base so scratch would be good, am I correct about this?

components/core/Dockerfile
kmehant: Done 😃!

components/core/Dockerfile
kmehant: @agentmilindu Can we get this merged?

agentmilindu: Oh, now I see! You are copying stuff into `scratch` image before you set the `WORKDIR`! Which means you are copying the root level files into `scratch as production` from `python:3.7-slim` not the Python files of Bassa, right? but don't it make `production` as same as `python:3.7-slim`? What's the use? I was wondering from where the `scratch` image got python3 binary!

kmehant: > Oh, now I see! You are copying stuff into `scratch` image before you set the `WORKDIR`! Which means you are copying the root level files into `scratch as production` from `python:3.7-slim` not the Python files of Bassa, right? but don't it make `production` as same as `python:3.7-slim`? What's the use? I was wondering from where the `scratch` image got python3 binary!

Yes @agentmilindu It does make as same as `python:3.7-slim`. I think this another way of doing this. We can have the base image as `python:3.7-slim` and copy just the **workdir** or else copy the whole container to scratch. Either should work fine 😃

kmehant: > Oh, now I see! You are copying stuff into `scratch` image before you set the `WORKDIR`! Which means you are copying the root level files into `scratch as production` from `python:3.7-slim` not the Python files of Bassa, right? but don't it make `production` as same as `python:3.7-slim`? What's the use? I was wondering from where the `scratch` image got python3 binary!

There seems no use between **base** stage and **production** stage. But for the sake of **linter** stage, I had to divide base and production so as to not add linting tools and related stuff to production. So if linting fails then the build fails there itself without moving to production 😄

agentmilindu: I prefer keeping `python:3.7-slim`. Because someone who already has `python:3.7-slim` image does not have to download the whole 143MB again then.

kmehant: I didn't get you @agentmilindu 😅. 

> I prefer keeping `python:3.7-slim`. Because someone who already has `python:3.7-slim` image does not have to download the whole 143MB again then.

kmehant: @vivonk @agentmilindu I have made the required changes. I have understood now Milindu I didn't think about the caching issue. Can we get this merged!

kmehant:
@vivonk @agentmilindu I didn't think about caching issue and now everything is fixed.
Can we get this merged now 😃!

components/core/Dockerfile
agentmilindu: Are you sure you pushed the changes? I still see we using `scratch` base image.

agentmilindu:
@kmehant Are you sure you pushed the changes? I still see we using `scratch` base image.

kmehant:
> @kmehant Are you sure you pushed the changes? I still see we using `scratch` base image.

@agentmilindu Sorry, I have pushed the changes now please review!

agentmilindu:
Why we do install dependencies twice? Let's do like this, 

FROM python-slim as base image
- copy stuff
- install dependencies
FROM base as lint 
- installs and runs linter
FROM base as production
- setup and run Bassa

kmehant:
@vivonk @agentmilindu I think we can merge this now 😄 
@agentmilindu Thank you! I was little carried away by container copying and made few mistakes and now I think I have come up with an enhanced fix for API Dockerfile. 
Please review and let's get this merged.

vivonk:
@agentmilindu can you please approve this one?

components/core/Dockerfile
```diff
@@ -1,32 +1,60 @@
...
 ARG VCS_REF
 
-WORKDIR /home/Bassa
-EXPOSE 5000
-COPY . .
+# Volume for accessing the downloaded files
+VOLUME [ "/downloads" ]
 
+# Setting current work directory to /home/Bassa
+WORKDIR /bassa
```
agentmilindu: This is not `/home/bassa` this is `/bassa` either change the comment or change the command

kmehant: @agentmilindu @vivonk Fixed :) !

agentmilindu: In Linux `/bassa` and `/Bassa` are two different things

components/core/Dockerfile
```diff
@@ -1,32 +1,60 @@
...
-&& ./setup.sh \
-# && python3 setup.py test \ disabing temporary
-&& python3 setup.py develop
-
-CMD [ "./run.sh"]
+# Exposing ports where the server listens 
+EXPOSE 5000
 
+# Start the Bassa API server
+CMD [ "python3", "Main.py" ]
```
JaDogg: If we use something like Gunicorn we can have few processes in the container 🤔

kmehant: @JaDogg bro that's a great idea. Shall we have a Gunicorn server for the production Dockerfile? Let's keep this Dockerfile as dev Dockerfile. Let me create a new PR with API integrated with WSGI and nginx. What are your thoughts about it? If yes please approve the PR 😄

JaDogg: 😂

JaDogg: Of course

Issue #781, opened 2019-05-21
Unable to spin up Bassa UI container for local development purpose


kmehant:
**Describe the bug**
Presently `development dockerfile` for Bassa UI is having issues in building into an image.

**To Reproduce**
* `Bassa UI dev dockerfile` resides as `Bassa/ui/Dockerfile.dev`.
* Try buidling an image from the above dockerfile

**Expected behavior**
On building the image from the local development dockerfile, Bassa UI container should successfully spin up.

kmehant:
Fixed!

Issue #375, opened 2018-01-01
Troubleshooting


agentmilindu:
I created a new wiki page for [Troubleshooting](https://github.com/scorelab/Bassa/wiki/Troubleshooting)

Let's tag this thread in all the installation and setting up errors, thus we can see the most common errors and their remedies, which we can add to the wiki easily.

agentmilindu:
Setting up Bassa in Windows https://github.com/scorelab/Bassa/issues/376

agentmilindu:
DB Connection issues https://github.com/scorelab/Bassa/issues/381

kmehant:
Error on running gulp serve with higher versions of node. https://github.com/scorelab/Bassa/issues/696

kmehant:
**Internal Server Error** when trying to login due to not setting up Bassa API server and MySQL properly and configurations not set properly.

VibhorCodecianGupta:
@kmehant error on running gulp-server is because of an issue with node-sass v4.5.0
In the release notes for that version, it says for OSX (and other OS as well I think) it supports upto node 7 and since ui requires a node engine >=6.10.x, anything besides this range will throw an error. 
@agentmilindu @vivonk I think this should be clarified in the documentation till the time the frontend codebase is not moved to react and webpack

kmehant:
Bassa is not compatible with higher versions of MySQL.
Possible solutions:
* Downgrade to MySQL 5.X
* [Change encryption method](https://stackoverflow.com/questions/49194719/authentication-plugin-caching-sha2-password-cannot-be-loaded)

hv7214:
In `dl.conf`, `aria_server` key's value(mainly hostname) has to changed according to the environment:

**Source-code setup:** ws://localhost:6800/jsonrpc
**Container setup:** ws://aria2c:6800/jsonrpc

**Note:** You need to run again `sudo python3 setup develop` command to overwrite the `~/.config/bassa/dl.conf` file.

PR #782, opened 2019-05-21
Updated and working version of development dockerfile for Bassa UI

kmehant:
## Description
Presently the `development Dockerfile` for Bassa UI is not properly written. We need a working  `dev Dockerfile` so that it satisfies the docker-compose and goes along with other components for local development.

## Related Issue

#781

## Types of changes

## Checklist:

### DIFF
--- a/ui/Dockerfile.dev
+++ b/ui/Dockerfile.dev
@@ -5,12 +5,16 @@ MAINTAINER SCoRe Lab Community <commuity@scorelab.org>
 ARG BUILD_DATE
 ARG VCS_REF
 
+
 WORKDIR /home/Bassa
 
+# importing complete ui component
 COPY . .
 
+# exposing necessary port
 EXPOSE 3000
 
+# installing all necessary packages for Bassa web
 RUN apt-get update \ 
 && apt-get install curl -y \
 && apt install nodejs npm -y \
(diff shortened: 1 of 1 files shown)


ui/Dockerfile.dev
```diff
@@ -1,4 +1,4 @@
-FROM node
+FROM ubuntu
```
JaDogg: What is the reason for using ubuntu image instead of docker?

kmehant: Hi @JaDogg bro, 
I had problems running with `node` as a base image so I have used `ubuntu` to install nodejs manually and it worked fine.

vivonk: which version of ubuntu this will pick? `FROM ubuntu` is it bionic or xenial. Please check that and if it's xenial then change it to `FROM ubuntu:18.04`

agentmilindu: I prefer to have to official NodeJs image. This way, when the official repo gets updated, we too get these updates. If you got issues with official repo, we can discuss those.

kmehant: @agentmilindu then let's have a specific version of nodejs base image (let's say 6) so that Bassa angular UI will be compatible. We had a discussion about breakdown due to higher versions of nodejs in this [thread](https://github.com/scorelab/Bassa/issues/375), What would you say? Shall we go this way?

ui/Dockerfile.dev
```diff
@@ -1,21 +1,36 @@
...
+# importing complete ui component
 COPY . .
 
-RUN apt-get update \
+# exposing necessary port
+EXPOSE 3000
+
+# installing all necessary packages for Bassa web
+RUN apt-get update \ 
+&& apt-get install curl -y \
```
JaDogg: Why do we need curl?

kmehant:
> Seems OK to me. But would be good if we can just use `node:version` like base image.

I have updated dockerfile accordingly Please review!

ui/Dockerfile.dev
kmehant: @agentmilindu I have fixed it. Please review :)

Issue #783, opened 2019-05-21
Improving installtiom scripts to update required installed packages


demon-36:
**Is your feature request related to a problem? Please describe.**
During Bassa installation for various OS flavors, after installing required packages it will be good to update them to latest

**Describe the solution you'd like**
Added commands for the packages updation

PR #784, opened 2019-05-21
Updated installation scripts

demon-36:
## Description
 
Updated installation scripts a bit for OS flavors that will update the required installed packages during Bassa installation

## Related Issue

#783 

## Types of changes

- [x] Improvement (non-breaking change which improves scripts)

## Checklist:

### DIFF
--- a/package-list-aptget
+++ b/package-list-aptget
@@ -1,5 +1,7 @@
 #!/bin/bash
-echo "Setting things to install Bassa..."
+echo "Preparing to install Bassa..."
+
+sudo apt-get update
 
 echo "Installing pip"
 sudo apt-get install python3-pip
@@ -33,5 +35,4 @@ echo "Installing gulp globally"
 sudo npm install -g gulp
 echo "gulp installed"
 
-sudo apt-get update
-echo "Finshed required components installation !"
+echo "Finshed installing required components !"
--- a/package-list-aptget
+++ b/package-list-aptget
@@ -1,4 +1,5 @@
 #!/bin/bash
+echo "Setting things to install Bassa..."
 
 echo "Installing pip"
 sudo apt-get install python3-pip
@@ -31,3 +32,6 @@ echo "bower installed"
 echo "Installing gulp globally"
 sudo npm install -g gulp
 echo "gulp installed"
+
+sudo apt-get update
+echo "Finshed required components installation !"
--- a/package-list-brew
+++ b/package-list-brew
@@ -1,7 +1,7 @@
 #!/bin/bash
-
+echo "Setting things to install Bassa..."
 echo "Installing Homebrew"
-/usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
+/usr/bin/ ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
 echo "Homebrew Installed"
 
 echo "Installing python3 and pip"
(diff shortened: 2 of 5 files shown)


package-list-aptget
```diff
@@ -1,4 +1,5 @@
 #!/bin/bash
+echo "Setting things to install Bassa..."
```
JaDogg:
```suggestion
echo "Preparing to install Bassa..."
```

package-list-aptget
```diff
@@ -31,3 +32,6 @@ echo "bower installed"
 echo "Installing gulp globally"
 sudo npm install -g gulp
 echo "gulp installed"
+
+sudo apt-get update
+echo "Finshed required components installation !"
```
JaDogg:
```suggestion
echo "Finished installing required components !"
```

demon-36:
> Please use suggestion, to all `echo` statements that require it.

@JaDogg Thanks for your suggestion. Actually each of `echo` statements are already written about describing the task they are going to perform :)

package-list-aptget
demon-36: @JaDogg bro I used `Setting things` as it will let the user know that this `echo` statement is about command that is going to install all required packages, So I found that more helpful :sweat_smile:

vivonk: I guess Preparing seems a good word @demon-36

vivonk: Can you change this to all scripts? @demon-36

package-list-brew
```diff
@@ -1,9 +1,13 @@
 #!/bin/bash
+echo "Setting things to install Bassa..."
 
 echo "Installing Homebrew"
-/usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
+/usr/bin/ ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
```
kmehant: This is broken due to the extra space before `ruby` in the url

kmehant: @demon-36 
Can you please fix that and **good practice would be to test it before sending a PR** 😀

package-list-brew
```diff
@@ -29,3 +33,6 @@ echo "Installing Gulp"
 npm install -g gulp
 npm install -g gulp-cli
 echo "Gulp Installed"
+
+brew update
```
kmehant: Can you please shift this step after brew doctor

package-list-dnf
```diff
@@ -45,3 +46,6 @@ echo "bower installed"
 echo "Installing gulp globally"
 npm install -g gulp
 echo "gulp installed"
+
+dnf check-update
```
kmehant: It's better to update before installing packages rather after.

kmehant: Can you please change the same for all the scripts

package-list-aptget
demon-36: okay @vivonk will do it for all.

package-list-dnf
demon-36: Actually I started updating scripts from CentOS  and that script was already having packages list updation before installing them that's how I missed it from other scripts
and it is recommended to update package-lists before installing packages so will do it for other OS scripts

package-list-brew
demon-36: @kmehant after going through ruby scripts syantax I think you are correct bro, earlier I thought this script would install Homebrew in mentioned directory(`/usr/bin`) using `ruby` as command parameter,

Issue #810, opened 2019-07-02
Connecting the application components with the help of React Router according to the workflow of Bassa.


singhrohan62:
**Is your feature request related to a problem? Please describe.**
After the completion of the static components implementation, it's time to connect the disjoint components and place them according to the flow of the application. For this to happen, I will be using React Router for adding routes to the application.

**Describe the solution you'd like**
For the Bassa React application, five total routes have been made. Out of five, four of them are protected routes, i.e., user cannot access those four routes without getting being authenticated. The one remaining route contains `Login` component which the user will be using for getting authenticated and hence, using the features of Bassa application.

**Additional context**
I would like to add a component for `404` route to the application. It would be fun and interesting to see that how the application responds to non-existing routes being queried by the user. I would explore such examples and draw inspiration from them and create our own component for 404 route!

PR #812, opened 2019-07-04
completed routing in the application and added GET requests for displaying completed and queued downloads

singhrohan62:
Completed route setup of the application and in components Queued, Component and Dashboard, made GET requests to the Bassa server to display completed and queued downloads.

## Description
* To add routes to the application, I added react-router-dom package to the app and created routes for the application.
* I created four private routes and one public route, this I have mentioned briefly in the issue #810 
* Later, with discussion with Milindu, I tried to make GET requests to the Bassa server and displayed the completed and queued downloads in the above mentioned 3 components.

I would like to create an another PR for adding all requests to Bassa server. Kindly allow me to do this as because I think that due to adding of GET requests in this PR, I have somewhat deviated from the issue I was working on. If we take small steps, it would be easy to track errors and find its solution.

Second thing, right now, for making those requests, I am handling the incoming data via state variables in the component. Later, when redux will be added to the app, I will be managing and handling data from the store with the help of reducers.

## Related Issue
Issue #810 

## Types of changes

## Checklist:


singhrohan62:
I would like to request everyone to review and provide feedback so that I can fix the issues if present in this PR and so that we can increase the pace of the project progress.

components/frontend-react/src/containers/Dashboard.js
```diff
@@ -58,6 +60,40 @@ const styles = theme => ({
...
+      completedList:[],
+      queuedList:[]
+    }
+  }
+
+  componentWillMount() {
+    let token = sessionStorage.getItem('token');
+    axios({
+      method:'get',
+      url: `${process.env.REACT_APP_API_URL}/api/user/downloads/1`,
```
vivonk: Getting complete URL from `process.env`, instead of hard coding endpoint definition here, makes maintain everything under single source easy:) 
LIke `process.env.READCT.. + process.env.API_USER_DOWNLOAD_1`

singhrohan62: Or maybe, Nirmal. we can add a list of all requests in env file and use them like,
```
REACT_APP_API_GET_DOWNLOADS=http://0.0.0.0:5000/api/user/downloads/1
REACT_APP_API_POST_LOGIN=http://0.0.0.0:5000/api/login
```
I think this one will be better. What are your thoughts?

agentmilindu: Let's keep these as is, anyway calling API endpoints should go inside Redux Sagas, this is a temporary state we have here.

iamchathu: Agreed @agentmilindu

Issue #816, opened 2019-07-19
Managing flow of states in the application with the help of Redux.


singhrohan62:
With complete routing setup of the application, we need to handle state and data present in the application.
I would like to use Redux along with Redux-Saga to handle API calls and http requests to the server.

Issue #814, opened 2019-07-08
Can't see the password


cjprogamer:
**There is no option for show password**

**Expected Behavior**
when we Signup/Signin then where should be an option for show password

**Current Behavior**
Can't see the password that I typed.

kmehant:
@cjprogamer thank you for reporting to us! Our angular UI is no longer being actively maintained as we are shifting to react UI this summer. So we will get this resolved in our new UI.

PR #817, opened 2019-07-20
Setup redux along with redux saga for handling asynchronous requests along with the data

singhrohan62:
In this pull request, I will be adding redux to the application along with its middleware redux-saga for handling AJAX requests to the Bassa server and handling incoming data.

## Description
First of all, I beg pardon for making changes in so many files, but the application which needed a redux setup along with its middleware took such number.

So, the state of the application was divided into two halves, one for handling user related requests and other for handling download (file) related request. You can check that I have used the same number of reducers, sagas and actions.

## Related Issue
I will be right now fixing issue #816 . For issue #814 , I would like to create a new PR since, the aim of this PR is to add redux along with redux saga. 

## How Has This Been Tested?
Testing of API calls were done for this PR.

## Types of changes

## Checklist:

- [] I have added tests to cover my changes.


components/frontend-react/src/API/Users.js
```diff
@@ -0,0 +1,55 @@
...
+  formData.set('user_name', action.creds.username);
+  formData.set('password', action.creds.password);
+  return axios.post(
+    `${process.env.REACT_APP_API_POST_LOGIN}`,
+    formData,    
+    {headers: {'Content-Type': 'multipart/form-data'}}
+  );
+}
+
+const addNewUser = (action) => {
```
agentmilindu: Doesn't `addNewUser` need a token?

components/frontend-react/src/App.js
```diff
@@ -1,31 +1,37 @@
...
+import { connect } from 'react-redux';
+
 import LoginComponent from './containers/LoginComponent';
 import AdminComponent from './containers/AdminComponent';
 import Dashboard from './containers/Dashboard';
 import QueuedComponent from './containers/QueuedComponent';
 import CompletedComponent from './containers/CompletedComponent';
 
-const App = () => (
+const App = (props) => (
```
agentmilindu: let's change this to 
```
const App = ({isloggedIn}) => (
```
Then you can use, 

```
<Route exact path='/' isloggedIn={isloggedIn} component={LoginComponent} />
```

components/frontend-react/src/API/Users.js
singhrohan62: Milindu, it's working without token being required. Actually, I took reference from Angular UI and found out that token was not required

components/frontend-react/src/components/UserSignup.js
```diff
@@ -42,19 +42,53 @@ const useStyles = makeStyles(theme => ({
...
+    }
+  }
+
   const handleSubmitButton = (event) => {
     event.preventDefault();
-    props.onClickSubmit();
+    let details = {user_name: username, email: email, password: password, confirm_password: confirmPassword};
+    if(password !== confirmPassword)
+    {
+      alert('Passwords do not match');
```
agentmilindu: I hope this is temporarily, later you will be replacing this with a model. 

I think it is better to add `TODO` comments,

```
// TODO Replace with a model dialog
alert('Passwords do not match');
```

singhrohan62: I thought of using react-toast component. I will add in my next PR

singhrohan62:
I just wanted to share something with you all. These are my plans for next PR I would like to send. 
1. I would try to add some new tests in the application. I will try to add tests for sagas.
2. I would like to add view for checking password.(For  issue #814)
3. I would like to add a component for `loading...` view
4. I would like to add a component for `404` route.

components/frontend-react/src/containers/AdminComponent.js
```diff
@@ -1,29 +1,29 @@
...
-    console.log('Kill downloads');
-  };
-  
   return (
     <div>
       <Appbar data-test="component-appbar" isloggedIn={true} />
-      <StartKillDownloads data-test="component-start-kill-downloads" onStart={startDownloads} onKill={killDownloads} />
-      <PendingSignupRequests data-test="component-pending-signup-requests" requestsList={requestsList}/>
-      <UsageStats data-test="component-usage-statistics" data={usageStatsData}/>
+      <StartKillDownloads data-test="component-start-kill-downloads" onStart={props.startDownloads} onKill={props.killDownloads} />
```
agentmilindu: Let's destruct the props at first? 

```
  const { startDownloads, killDownloads } = props;
  return (
  ...
````

Comment (agentmilindu) [reply components/frontend-react/src/containers/AdminComponent.js]: @iamchathu @JaDogg @vivonk WDYT?

File: components/frontend-react/src/containers/Dashboard.js [hunk 5]
```diff
@@ -61,27 +63,25 @@ const styles = theme => ({
 
 class Dashboard extends React.Component {
 
-  constructor(){
-    super();
+  constructor(props){
+    super(props);
```
agentmilindu: A fancy way of doing this is
```
class Dashboard extends React.Component {
  state = {
    link: '',
    completedList:[]
  }
```
and then remove the constructor.

I think I remember it correctly, please correct me if I'm wrong. @iamchathu .

components/frontend-react/src/containers/Dashboard.js
```diff
@@ -61,27 +63,25 @@ const styles = theme => ({
...
+      method: 'get',
+      url: `${process.env.REACT_APP_API_GET_DOWNLOADS}`,
       headers: {'token': `${token}`}
     })
-    .then( res => {
-      let completedList = res.data.filter(file => file.status === 3);
+    .then(res => {
       let queuedList = res.data.filter(file => file.status === 0);
-      this.setState({completedList, queuedList})
+      let completedList = res.data.filter(file => file.status === 3);
```
agentmilindu: let's not have magic numbers, define these number as constants somewhere and then use. 
```
const PENDING = 2;
const DOWNLOADING = 3;

...

.then(res => {
   let completedList = res.data.filter(file => file.status === DOWNLOADING);
``

File: components/frontend-react/src/containers/Dashboard.js [hunk 7]
```diff
@@ -94,21 +94,30 @@ class Dashboard extends React.Component {
...
+  	  	  Hi {this.props.username}!
   	  	</Typography>
   	    <Paper className={classes.paperAdd}>
   	      <Typography variant="h5">
   	        ADD DOWNLOAD
   	      </Typography>
           <form>
-            <Input type="text" className={classes.link} placeholder="Enter or paste the link below" />
-            <FAB color="primary" size="small" className={classes.fab} aria-label="add-download">
+            <Input type="text" className={classes.link} value={this.state.link} onChange={this.handleLinkField} placeholder="Enter or paste the link below" />
```
agentmilindu: let's destruct state also somewhere above,

```
const { link } =  this.state;
````

**Comment by agentmilindu** [comment]
Let's change the `API` directory into `service`?

File: components/frontend-react/src/API/Users.js (hunk 1 above)
Comment (agentmilindu) [reply components/frontend-react/src/API/Users.js]: @vivonk Is this the expected behaviour?

File: components/frontend-react/src/components/UserSignup.js (hunk 3 above)
Comment (agentmilindu) [reply components/frontend-react/src/components/UserSignup.js]: Cool

File: components/frontend-react/src/API/Users.js (hunk 1 above)
Comment (singhrohan62) [reply components/frontend-react/src/API/Users.js]: Milindu, what I understood is that the first time user won't  be able to access token as it is generated only when user gets logged in. So, in order to get registered he needs to submit the form for getting approved by the admin.

File: components/frontend-react/src/containers/AdminComponent.js (hunk 4 above)
Comment (iamchathu) [reply components/frontend-react/src/containers/AdminComponent.js]: @agentmilindu Agreed. Shall we add some `eslint` rules to enforce this ?

File: components/frontend-react/src/containers/Dashboard.js (hunk 5 above)
Comment (iamchathu) [reply components/frontend-react/src/containers/Dashboard.js]: @agentmilindu With `classProperties` syntax support now we don't want to use constructor to initialize state.

File: components/frontend-react/src/API/Downloads.js [hunk 8]
```diff
@@ -0,0 +1,50 @@
+import axios from 'axios';
+
+const fetchDownloads = () => {
+  let token = sessionStorage.getItem('token');
```
iamchathu: We can use `axios` `interceptors` to inject token for the requests. @agentmilindu what do you think?

components/frontend-react/src/API/Downloads.js
```diff
@@ -0,0 +1,50 @@
...
+
+const killAllDownloads = () => {
+  return axios.get(
+    `${process.env.REACT_APP_API_KILL_DOWNLOADS}`,
+    {headers: {'key': '123456789'}}
+  );
+}
+
+const addNewDownload = (link) => {
+  let obj = {link: link};
```
iamchathu: `let reqData = {link} `

Since the Object property shorthand syntax we don't need to repeat `{link:link}` if the variable name and property name is equal.

components/frontend-react/src/App.js
iamchathu: Yeah every time at the beginning of the function or render distrust the state and props before use.

components/frontend-react/src/components/Appbar.js
```diff
@@ -62,6 +64,7 @@ const BassaAppBar = (props) => {
     switch(route){
       case 'logout':
         sessionStorage.removeItem('token');
+        props.logoutUser()
         props.history.push('/');
```
iamchathu: You need to clean the history and the push the basepath. Try replace. Read the history documentation.

components/frontend-react/src/components/UserSignup.js
```diff
@@ -42,19 +42,53 @@ const useStyles = makeStyles(theme => ({
 
 const UserSignup = (props) => {
 
+  const [username, setUsername] = useState('');
+  const [email, setEmail] = useState('');
+  const [password, setPassword] = useState('');
+  const [confirmPassword, setconfirmPassword] = useState('');
+
+  const handleTextFields = (event) => {
```
iamchathu: @agentmilindu Shall we use `formik` to handle form data and validations?

components/frontend-react/src/API/Downloads.js
singhrohan62: I will make the necessary changes

components/frontend-react/src/containers/AdminComponent.js
```diff
@@ -1,29 +1,29 @@
...
+      <PendingSignupRequests data-test="component-pending-signup-requests" requestsList={props.requestsList}/>
+      <UsageStats data-test="component-usage-statistics" data={props.usageStats}/>
     </div>
   )
 }
 
-export default AdminComponent;
\ No newline at end of file
+const mapStateToProps = state => ({
+  requestsList: state.userReducer.requests,
```
iamchathu: Shall we right a memoized selector for this? @agentmilindu WDYT?

components/frontend-react/src/containers/QueuedFileList.js
```diff
@@ -1,5 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+import { deleteDownload } from '../actions/downloadActions';
 
 //MUI imports
 import { withStyles } from '@material-ui/core/styles';
```
iamchathu: Use makeStyles hooks way or withStyles HOC only. I think you have used makeStyles in some places, right?

components/frontend-react/src/index.js
```diff
@@ -1,11 +1,15 @@
...
+import { Provider } from 'react-redux';
+import {configureStore } from './getStore';
 import './index.css';
 import App from './App';
 import * as serviceWorker from './serviceWorker';
 
-ReactDOM.render(<BrowserRouter><App/></BrowserRouter>, document.getElementById('root'));
+const store = configureStore();
+
+ReactDOM.render(<Provider store={store}><BrowserRouter><App/></BrowserRouter></Provider>, document.getElementById('root'));
```
iamchathu: Extract this to functional component above and then use in here.

`const Application = ()=> (<Provider>... )`

components/frontend-react/src/reducers/downloadReducer.js
```diff
@@ -0,0 +1,37 @@
...
+const initialState = {
+  completedDownloads: [],
+  queuedDownloads: []
+};
+
+const userReducer = (state = initialState, action) => {
+  switch(action.type) {
+
+    case downloadActions.FETCH_DOWNLOADS:
+      let completedList = action.data.filter(file => file.status === 3);
```
iamchathu: Extract the numbers to constants and use

components/frontend-react/src/sagas/downloadSaga.js
```diff
@@ -0,0 +1,49 @@
+import { take, takeEvery, put, call } from 'redux-saga/effects';
+import { startAllDownloads, killAllDownloads, addNewDownload, fetchDownloads, deleteDownload } from '../API/Downloads';
+import { downloadActions } from '../constants';
+
+function* startAllDownloadsWorkerSaga () {
+  yield call(startAllDownloads);
```
iamchathu: Sagas should handle it's effects. Use try catch . What if your requests fails?  @agentmilindu WDYT?

components/frontend-react/src/sagas/userSaga.js
```diff
@@ -0,0 +1,46 @@
+import { takeLatest, put, call, take } from 'redux-saga/effects';
+import { userActions, downloadActions } from '../constants';
+import { verifyUserCredentials, addNewUser, getSignupRequests, approveSignupRequest, getHeavyUsers } from '../API/Users';
+import { fetchDownloads,  } from '../API/Downloads';
+
+function* userLoginWorkerSaga (action) {
+  const response = yield call(verifyUserCredentials, action);
+  sessionStorage.setItem('token', response.headers.token);
```
iamchathu: What is token is not there in `response`?

components/frontend-react/src/containers/QueuedFileList.js
singhrohan62: Yes Chathu, I have used withStyles for class components and makeStyles for functional components. As makeStyles is a React Hook while withStyles is a HOC, so I was unable to use makeStyles in class componennts.

components/frontend-react/src/sagas/userSaga.js
singhrohan62: I could not understand the question Chathu. Can you please explain it to me?

iamchathu: @singhrohan62 Sorry if I confuse you. What if `response.headers.token` is undefined?

components/frontend-react/src/components/UserSignup.js
agentmilindu: yeah, that sounds good

components/frontend-react/src/containers/AdminComponent.js
agentmilindu: agree!

components/frontend-react/src/sagas/downloadSaga.js
agentmilindu: Agree with you. The failure scenario also should be considered.

components/frontend-react/src/API/Downloads.js
agentmilindu: I was thinking the same. totally agree.

components/frontend-react/src/containers/AdminComponent.js
agentmilindu: Agree, Let's have `eslint` rule for this. @singhrohan62

components/frontend-react/src/sagas/userSaga.js
singhrohan62: Chathu, I had thought about this, and came up with an idea that when we receive token from response, we can issue another GET request in order to check whether token is valid or not. Do you think it's a good idea ?

agentmilindu: No, as I remember, the tokens are always generated with each request, like a rolling token. @vivonk  correct me if I'm wrong.

 And making API call for each request is also not good. Just put validation to check if `response` or `response.headers` is there. If they ever are `undefined`, that will cause an exception and the app might crash.

@vivonk

singhrohan62:
Hi everyone, just needed to tell you that I would like to add eslint configuration for destructuring the state and props of the component in the next commit. Also, kindly allow me to address Chathu's suggestions in incoming later commits. Thank you!

components/frontend-react/src/containers/Dashboard.js
```diff
@@ -63,22 +66,22 @@ const styles = theme => ({
...
     let token = sessionStorage.getItem('token');
     axios({
       method: 'get',
       url: `${process.env.REACT_APP_API_GET_DOWNLOADS}`,
       headers: {'token': `${token}`}
     })
     .then(res => {
-      let queuedList = res.data.filter(file => file.status === 0);
-      let completedList = res.data.filter(file => file.status === 3);
+      let queuedList = res.data.filter(file => file.status === QUEUED);
```
JaDogg: Nice

singhrohan62:
Hi everyone! I made changes according to Chathu's comments he told and added component for '404' route. Also, I fixed issues #822 and #814

singhrohan62:
Screenshot from 2019-08-19 20-04-46
Password field now has the visibility option to check the password typed.

Screenshot from 2019-08-19 20-05-04
When user enters a route which is not defined in the application.

agentmilindu:
@JaDogg @iamchathu Shall we review this and merge?

components/frontend-react/src/containers/LoginComponent.js
```diff
@@ -1,49 +1,78 @@
...
+
+const validationSchema = Yup.object({
+  user_name: Yup.string("Enter a name").required("Name is required"),
+  email: Yup.string("Enter your email")
+    .email("Enter a valid email")
+    .required("Email is required"),
+  password: Yup.string("")
+    .min(8, "Password must contain atleast 8 characters")
+    .required("Enter your password"),
+  confirm_password: Yup.string("Enter your password")
```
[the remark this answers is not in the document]
singhrohan62: Hi @JaDogg I changed it to confirmPassword.

components/frontend-react/src/components/UsageStats.js
```diff
@@ -16,27 +14,77 @@ const useStyles = makeStyles(theme => ({
   chart: {
     paddingLeft: 30,
   }
-}))
+}));
+
+const renderActiveShape = (props) => {
+  const RADIAN = Math.PI / 180;
```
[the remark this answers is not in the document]
singhrohan62: I added comments representing the usage of variables

PR #818, opened 2019-07-21
Design revamp

VibhorCodecianGupta:
## Description
Implemented Composite design pattern with two constructs only, namely Folders and Files. The differentiation of projects and workspaces will be handled at the user end rather than the backend, to keep things simple and database management easy.
This allows us to very easily manage ACL flows as we have essentially a single entity to deal with, which is the `EntityInterface` abstract class, with Folder and File as its derived classes.
This also helps in writing recursive sql and operations, since we only have to care about a single data structure and hence less costly operations.

## Motivation and Context
To simplify and deconstruct the complex architecture previously thought out. Design pattern allows more flexibility and less, concise code

## How Has This Been Tested?
Mock tested functions and integration tested APIs using Postman

## Types of changes

## Checklist:


VibhorCodecianGupta:
**TODO**
 - [x] Models
 - [x] SQL scripts
 - [x] APIs and helper methods
 - [x] ACL flow
 - [x] File upload
 - [x] API tests
 - [ ] UI for File Management System

VibhorCodecianGupta:
@agentmilindu @vivonk, please review :)

components/core/routes/Directory.py
```diff
@@ -0,0 +1,131 @@
...
+    resp.headers['Access-Control-Expose-Headers'] = 'token'
+    return resp
+
+
+def fetch_entity(id):
+    token = token_validator(request.headers.get('token'))
+    if token is None:
+        return '{"error":"token error"}', 403
+    try:
+        type = request.args.get('type')
```
JaDogg: Rename `type` to something else as it's a python builtin. https://docs.python.org/3.5/library/functions.html#type

components/core/utils/test_method_utils.py
```diff
@@ -0,0 +1,53 @@
+from managers.DirectoryManager import Folder, File
+
+
+def username():
```
JaDogg: Why are there methods that return a single value?

VibhorCodecianGupta: @JaDogg this kind of convention was being used already for functional tests to pass parameters to functions, so I continued with it. Of course, we can remove this and directly pass hard values as well. Thoughts?

components/core/routes/Directory.py
VibhorCodecianGupta: done

components/core/routes/Acl.py
```diff
@@ -0,0 +1,47 @@
...
+    try:
+        entity = request.args.get('type')
+        if not entity == 'fr' and not entity =='fl':
+            return Response('{"error":"invalid parameter"}', status=400)
+
+        user_name = request.form['user_name']
+        access = request.form['access']
+
+        grant_response = give_access(id, user_name, entity, access)
+        resp = Response(response='{"status":"' + grant_response + '"}',
```
vivonk: I guess 500 is not a suitable code for permission error, 4xx? @VibhorCodecianGupta

components/core/routes/Upload.py
```diff
@@ -0,0 +1,42 @@
...
+        upload_response = upload_file_helper(form_field)
+
+        if not 'msg' in upload_response:
+            resp = Response(response='{"error":"'+ upload_response['err'] +'"}', status=500)
+
+        if upload_response['msg'] is not 'success':
+            resp = Response(response='{"error":"' + upload_response['msg'] + '"}', status=400)
+
+        file_body = "name="+upload_response['name']+"&parent_id="+parent_id
+        add_response = requests.post('http://localhost:5000/api/user/drive/'+user_id+'/add', file_body, params=e_type, headers=add_headers)
```
vivonk: can you change the URL to little dynamic as endpoint definition and hostname should be kept in constants file and then use over codebase @VibhorCodecianGupta

VibhorCodecianGupta: oh yes I intend to do that, should i make a config file for the host and post variables then? Correct me if i'm wrong, i don't think there is one at present

components/core/routes/Acl.py
VibhorCodecianGupta: this is not a permission error, this is made in a way that 4xx errors will be either token error 403 (forbidden), a 400 (bad request) code if the query param is invalid or a 400 when some exception is thrown otherwise. This line gets the response from the function that hits the db, so it's either 200 or 500 server error if something goes wrong. the 4xx errors are all handled prior to this line. thoughts?

components/core/routes/Upload.py
vivonk: Ya, config might be the right choice. Go for it :)

components/core/routes/Acl.py
vivonk: If the user doesn't have access, it doesn' mean `error` I think. 403 (forbidden) shows you can't perform this action. 
Let's say a user has correct token but it doesn't belong to the user who is trying to perform this action then it's a Forbidden state I think 
@agentmilindu @JaDogg what do you think?

components/core/routes/Upload.py
VibhorCodecianGupta: cool, i'll do that

components/core/routes/Acl.py
VibhorCodecianGupta: That's totally correct. I'm talking about the 500 code you commented on, I mean that because all the 4xx are handled before this 500 code, and because the code reaches here only if the user inputs are correct and the user has access, this method has already bypassed the 4xx errors. the only errors possible here are server errors if the method doesn't work, for example, db connection times out. which is a 500 error

JaDogg: This is what I use/recommend:
* 403 - any permission issue. Forbidden access - https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403
* 422 - input is parsed (read as JSON) but it's wrong - ex: fields are wrong, invalid data type, invalid enum (fr/fl in this case) https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422
* 500 - Avoid responding with a 500 intentionally. This is reserved for situations where there is some error on server side. (Unexpected exception, etc.)  https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/500

components/core/routes/Acl.py
```diff
@@ -0,0 +1,47 @@
...
+    token = token_validator(request.headers.get('token'))
+    if token is None:
+        return '{"error":"token error"}', 403
+    try:
+        entity = request.args.get('type')
+        if not entity == 'fr' and not entity =='fl':
+            return Response('{"error":"invalid parameter"}', status=400)
+
+        check_response = get_access(id, user_id, entity)
+        if not isinstance(check_response, str):
```
JaDogg: We can probably replace this with an exception. Raise an exception for an error instead of returning a string.

components/core/utils/test_method_utils.py
JaDogg: Why not just constants @vivonk 🤔

JaDogg: You can just use these values in the test itself I think.

VibhorCodecianGupta:
`drive.html` and upload functionality are WIP

components/core/routes/Acl.py
VibhorCodecianGupta: will do

VibhorCodecianGupta:
@vivonk any reviews? 
Here is the wiki: https://github.com/scorelab/Bassa/wiki/File-Management-System
Here is the endpoint doc: https://github.com/scorelab/Bassa/wiki/URL-Endpoints#storage-entity-endpoints

Issue #769, opened 2019-05-10
Make Bassa ready for Kubernetes deployments


kmehant:
**Is your feature request related to a problem? Please describe.**
Write now Bassa has been completely containerized using Docker and the docker-compose tool has been integrated for easy local deployments. Now we can go step further by deploying Bassa to a Kubernetes cluster for better orchestration and cloud deployments.

**Describe the solution you'd like**
I would like to divide the manifests according to the decided architecture for Kubernetes deployments.

hv7214:
@kmehant are you still assigned on this ?

kmehant:
Please look through the referenced pull requests and most of the work is done which might lead to duplication of work. Anyways you can look through #828.

hv7214:
@kmehant i see, is there any minor work left ?

kmehant:
Fixed :)

PR #828, opened 2019-09-02
Make Bassa ready for Kubernetes deployments

kmehant:
## Description
Written kubernetes configurations for Bassa API server, Bassa Angular client, database and aria2c. 

## Overview 
Web and Bassa database will be in separate pods, whereas aria2c and API server will be in a single pod sharing a common persistent volume for downloaded files. Bassa database uses another persistent volume for the database. Coming to hostnames, as Bassa API server and Aria2c server stays in the same pod so they communicate using the localhost hostname. API server communicates to the database through the hostname `db` which is being resolved by Kubernetes DNS. There is a problem for communication between Bassa UI and Bassa API server and presently shall use ingress for Bassa Web client to communicate with API server. Coming to persistent volumes which have to be replaced with other cloud-based options like AWS, GCE, and others. Need some suggestion to find a best way for communication between API and Web.

## Related Issue
#769 

## Motivation and Context
We need to integrate the orchestration tool Kubernetes to Bassa for better on-premise deployments.

## Types of changes

## Checklist:

### DIFF
--- a/components/core/Killer.py
+++ b/components/core/Killer.py
@@ -1,6 +1,6 @@
 import urllib.request, urllib.error, urllib.parse
 
-req = urllib.request.Request('http://localhost:5000/download/kill')
+req = urllib.request.Request('https://192.168.64.9/api/download/kill')
 req.add_header('key', '123456789')
 resp = urllib.request.urlopen(req)
 content = resp.read()
--- a/components/core/Killer.py
+++ b/components/core/Killer.py
@@ -1,6 +1,6 @@
 import urllib.request, urllib.error, urllib.parse
 
-req = urllib.request.Request('https://192.168.64.9/api/download/kill')
+req = urllib.request.Request('http://localhost:5000/download/kill')	
 req.add_header('key', '123456789')
 resp = urllib.request.urlopen(req)
 content = resp.read()
--- a/components/core/Server.py
+++ b/components/core/Server.py
@@ -27,8 +27,8 @@ def serve_ui1():
 
 
 # download endpoints
-server.add_url_rule(rule='/download/start', endpoint='start', view_func=Download.start)
-server.add_url_rule(rule='/download/kill', endpoint='kill', view_func=Download.kill)
+server.add_url_rule(rule='/api/download/start', endpoint='start', view_func=Download.start)
+server.add_url_rule(rule='/api/download/kill', endpoint='kill', view_func=Download.kill)
 server.add_url_rule(rule='/api/download', endpoint='add_download_request', view_func=Download.add_download_request,
 					methods=['POST'])
 server.add_url_rule(rule='/api/download/<int:id>', endpoint='remove_download_request',
--- a/components/core/dl.conf
+++ b/components/core/dl.conf
@@ -1,1 +1,1 @@
-{"max_downloads": 2, "down_folder": "../../downloads", "size_limit":100000, "max_age":1, "min_rating":3, "max_bandwidth": 300000, "aria_server": "ws://aria2c:6800/jsonrpc", "tmp_folder": "/tmp/bassa/"}
+{"max_downloads": 2, "down_folder": "../../downloads", "size_limit":100000, "max_age":1, "min_rating":3, "max_bandwidth": 300000, "aria_server": "ws://localhost:6800/jsonrpc", "tmp_folder": "/tmp/bassa/"}
--- a/components/database/dockerfile
+++ b/components/database/dockerfile
@@ -1,0 +1,5 @@
+FROM mysql:5.7
+COPY ../../db_schema/Bassa.sql /tmp/schema/
+COPY ../../.docker/volumes/mysql/create-schema.sh /docker-entrypoint-initdb.d/
+EXPOSE 3306
+CMD ["mysqld"]
--- a/scripts/kubernetes/deploy.sh
+++ b/scripts/kubernetes/deploy.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+
+if [ "$1" = "delete" ]; then
+    echo "Clearing the cluster."
+    kubectl delete -f ./manifests/web/00-namespace.yml
+    echo "Bassa is removed from the cluster"
+elif [ "$1" = "create" ]; then
+    echo "Deploying Bassa to kubernetes cluster"
+    # Deploy web pod
+    kubectl create -R -f ./manifests/web/
+    # Deploy api and aria2c pod
+    kubectl create -R -f ./manifests/api_and_aria2c/
+    # Deploy database pod
+    kubectl create -R -f ./manifests/database/
+    echo "Bassa got deployed at"
+    kubectl get ingress -n bassa | awk '{ print $3 }'
+fi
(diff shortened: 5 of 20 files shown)


kmehant:
@agentmilindu @JaDogg @vivonk Please review this, perhaps need suggestions in resolving communication between API and Web

components/core/Killer.py
```diff
@@ -1,6 +1,6 @@
 import urllib.request, urllib.error, urllib.parse
 
-req = urllib.request.Request('http://localhost:5000/download/kill')
+req = urllib.request.Request('https://192.168.64.9/api/download/kill')
```
kmehant: That IP address is my local ingress address, please suggest upon this.

components/core/Server.py
```diff
@@ -27,8 +27,8 @@ def serve_ui1():
 
 
 # download endpoints
-server.add_url_rule(rule='/download/start', endpoint='start', view_func=Download.start)
-server.add_url_rule(rule='/download/kill', endpoint='kill', view_func=Download.kill)
+server.add_url_rule(rule='/api/download/start', endpoint='start', view_func=Download.start)
+server.add_url_rule(rule='/api/download/kill', endpoint='kill', view_func=Download.kill)
```
kmehant: When we use ingress to route to Bassa services, we need `/api` in our endpoints to differentiate between client and api

components/core/dl.conf
```diff
@@ -1 +1 @@
-{"max_downloads": 2, "down_folder": "../../downloads", "size_limit":100000, "max_age":1, "min_rating":3, "max_bandwidth": 300000, "aria_server": "ws://aria2c:6800/jsonrpc", "tmp_folder": "/tmp/bassa/"}
+{"max_downloads": 2, "down_folder": "../../downloads", "size_limit":100000, "max_age":1, "min_rating":3, "max_bandwidth": 300000, "aria_server": "ws://localhost:6800/jsonrpc", "tmp_folder": "/tmp/bassa/"}
```
kmehant: Changed it localhost as Bassa API server and Aria2c reside in a single pod, therefore, they can use localhost for communication.

components/database/dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM mysql:5.7
```
kmehant: I had to write a Dockerfile for database to prevent injecting schema which needs extra volume to store the schema and db init shell script.

scripts/kubernetes/deploy.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
```
kmehant: Scripted an easy to deploy script for kubernetes

ui/src/app/index.js
```diff
@@ -3,7 +3,7 @@
 angular.module('bassa', ['ngAnimate', 'ngCookies', 'ngTouch',
   'ngSanitize', 'ui.router', 'ngMaterial', 'nvd3', 'app'])
 
-  .value('BassaUrl', 'http://localhost:5000')
+  .value('BassaUrl', 'https://192.168.64.9')
```
kmehant: For the web to contact API server through ingress

components/database/dockerfile
demon-36: @kmehant can't we directly use mysql docker-image from dockerhub ?

demon-36: @kmehant can you please describe why we need to write Dockerfile as we can directly pull mysql docker-image from dockerhub.

kmehant: @demon-36 we can do that also, as I have said earlier we need extra volumes for mounting schema and init db script. It better to bind them in to a separate image and use in k8s

ui/src/app/index.js
vivonk: This will break out existing APIs as a lot of APIs using `BassaUrl` value

kmehant: @vivonk bro, true but not sure how to handle this. Either we should manually change this to production Bassa API url when have deployed completely. Or we should use environment variables. I am unable to use environment variables here.

JaDogg: Is there anyway to use a config file?

kmehant: @JaDogg bro true, we can use a config file. On total I feel we have two options
1. Get values using environment variables (this makes the deployment very easy as we can feasibly change API url anytime without rebuilding the image) 
2. Use a common config file. We will mount that rather copying it into container, doing this will result in similar benefit as of using env variables. But we might need to have an extra volume to store the file. 
@JaDogg @vivonk @agentmilindu any thoughts about this?

JaDogg: Get value from somewhere and default to localhost?

components/core/Killer.py
```diff
@@ -1,6 +1,6 @@
 import urllib.request, urllib.error, urllib.parse
```
kmehant: @vivonk bro Are we using killer and starter python scripts 🤔 ?

ui/src/app/index.js
```diff
@@ -1,9 +1,10 @@
 'use strict';
+import { config } from "../../config";
 
 angular.module('bassa', ['ngAnimate', 'ngCookies', 'ngTouch',
   'ngSanitize', 'ui.router', 'ngMaterial', 'nvd3', 'app'])
 
-  .value('BassaUrl', 'http://localhost:5000')
+  .value('BassaUrl', config.API_URL)
```
kmehant: @vivonk @JaDogg I have made this change

JaDogg: Awesome

sanjulamadurapperuma:
@kmehant there is a conflicting file, could you resolve those and add another commit?

JaDogg:
Resolve conflict?

kmehant:
@JaDogg @sanjulamadurapperuma 
Hello, I have resolved the conflicts :)

sanjulamadurapperuma:
@kmehant did you check the codacy issues?

ui/src/app/admin/services/AdminService.js
```diff
@@ -8,15 +8,15 @@
     var startDownloads = function() {
       return $http({
           method: 'GET',
-          url: BassaUrl + '/download/start',
+          url: BassaUrl + '/api/download/start',
```
vivonk, line 11: Does we have backend changed for it as well? @kmehant Any idea?

vivonk: Because I guess `/api/` will be appended to actual request from ingress as well?

kmehant: @vivonk 
We have the server routes in `/api/` form and two routes belonging to start and kill downloads were also changed to that form. Server endpoints now all are in `/api/` form so in the same sense I have changed them in client-side too.

In the case of ingress the host is the same so had to add extension `/api/` to differentiate between client and server service.

vivonk:
@kmehant @sanjulamadurapperuma I guess codacy errors are going to be with repo till we make whole codebase to be either using linting or make pre-lints.

Ignoring Codacy error for this PR.

PR #841, opened 2019-12-23
Added  the Downloads folder

cjprogamer:


.gitignore
```diff
@@ -226,4 +224,13 @@ tags
 !.vscode/extensions.json
 .history
 
+<<<<<<< HEAD
```
kmehant: Fix these conflicts

.gitignore
```diff
@@ -226,4 +224,13 @@ tags
...
+###make empty downloads folder
+downloads/*
+!downloads/.gitkeep 
+=======
+downloads/*
+!downloads/.gitkeep
```
kmehant: Can you explain the reason .`gitkeep` is used here 
Can I use something like `.downloads` or something else?
Put forward your explanation.

kmehant:
Squash your commits

kmehant:
Work on `develop` branch not on master

.gitignore
cjprogamer: Is it correct if I add these lines ? 
`#downloads
.downloads`

kmehant: That results In a comment due to #
> Can you explain the reason .gitkeep is used here
Can I use something like .downloads or something else?

Hopefully you will answer this

cjprogamer: no, I mean,
" #downlaods
    .downloads

cjprogamer: > That results In a comment due to #
> 
> > Can you explain the reason .gitkeep is used here
> > Can I use something like .downloads or something else?
> 
> Hopefully you will answer this

I thought that code would not delete the downloads folder

kmehant:
#843 shifted

PR #845, opened 2019-12-24
Add alternative key method

JosephSemrai:
Changed config file to get the server secret key from an environment variable.

## Description
Used the `os` module in the Bassa core config Python file to grab the `BASSA_SERVER_KEY` environment variable.

## Motivation and Context
This solves a major problem with security as most configurations and instances of Bassa might not change this default SERVER_SECRET_KEY as the project will work without this configuration since it is hard coded.

## How Has This Been Tested?
Ran the project.

## Types of changes

## Checklist:


components/core/utils/app_constants.py
```diff
@@ -1 +1,2 @@
-SERVER_SECRET_KEY = "123456789"
+import os
+SERVER_SECRET_KEY = os.environ['BASSA_SERVER_KEY']
```
kmehant: `os.environ.get`
I encourage to use this here instead

kmehant:
Please squash your commits :)

Issue #719, opened 2019-03-25
Shall we avoid editing DBCon.py?


xlight05:
## Context
When we are configuring the database, we are encouraging people to use hard-coded values in DBCon.py.

I don't really mind hard coding values in a config file for testing purposes however it really is a bad practice to edit the code. I've done this myself and I accidentally staged DBCon.py couple of times. 

I think the solution is already implemented. In the default code, we are getting values from the environment variables. I believe its correct, I think we should encourage users to use the env variables and remove editing DBCon.py in the README and Wiki pages. 

Maybe we can provide a template .env file for usability. What do you think? If you agree,  I can start working on this.

hv7214:
@kmehant what you say? I think it's better to make a template .env file rather making env variables/directly populating credentials.. It becomes really easy for a developer to configure a database, and hence prevent accidental staging of DBCon.py file.

JaDogg:
Can we create an inheritable config system. We have `config.json` and `user.config.json`. We will commit `config.json` and not commit user config. We can put default values there in `config.json`.

When loading the config we load both and merge in a way `user.config.json` overrides `config.json` data. 

If we do `.ini` or `.yml` We can also have comments 😀
--

Then we create a config class to handle all this. Have every other class create an use configs or use a method like `cfg["config"]` (using` __getitem__`)

hv7214:
@JaDogg it seems a great idea to go, though i have a lil'bit different approach. We can have a `config.example.json`, which should be treated as a template file. User has to **duplicate** this file as `config.json` (leading to two files: `config.json` and `config.example.json`). So data will be loaded directly from `config.json` file. No merging stuff. And `config.json` will be added to `.gitignore`, so no headache of staging.

JaDogg:
We can simply use `ChainMap` for merging and some custom code to handle nested merging. We can use `.yml` and include comments and examples in the same config. Example config is OK as well for Bassa. 

Currently I'm not sure if it is possible to just run Docker to get stuff working. If I remember correctly we need to set some environment variables. @kmehant knows I think

hv7214:
as much i know env variables are used for database config.

kmehant:
> We can simply use `ChainMap` for merging and some custom code to handle nested merging. We can use `.yml` and include comments and examples in the same config. Example config is OK as well for Bassa.
> 
> Currently I'm not sure if it is possible to just run Docker to get stuff working. If I remember correctly we need to set some environment variables. @kmehant knows I think

[...]

PR #884, opened 2020-02-07
Issue719

hv7214:
## Description

- Made a `db.config.json` file which stores default values for db credentials.
- User can add `user.db.config.json` to override any of the credential.
- Added doc for same (*ADD_TO_WIKi.md*).
- Added **user.config.json** in .`gitignore`.
## Related Issue
#719

## Motivation and Context

Setting credentials with .env file is really a bad option, as it creates problem at the time of deployement(k8). And hardcoding credentials bring staging problem.

## How Has This Been Tested?

After change, db is successfully set up and can do usual operations(login etc.)

## Types of changes

## Checklist:

- [x] I have updated the documentation accordingly.

**Note**
This PR contains a `ADD_TO_WIKI.md` whose contents should be appended to db section of wiki accordingly.


components/core/DBCon.py
```diff
@@ -1,13 +1,15 @@
 import MySQLdb
+from ConfReader import get_conf_reader
 import os
 
 _db=None
+dbConf = get_conf_reader("db.config.json")
```
JaDogg: use snake case

hv7214: yeah sure

hv7214:
@JaDogg can you review this PR ? i have made proposed changes.

kmehant:
Having this setup has pros and cons. I guess you can find the pros here #719. Yet the cons are little but this change makes the container base testing or production setup pretty much tedious. for example, I need to rebuild huge images and spin up a container for as small as any of the changes in the credentials, that would create lag and latency.

That's one of the main reasons for using environmental variables for the container-based setup I can simply change the environmental variables and just restarting the containers will do the job.  And this is same for the normal setup you do :)

Let's not change this present model for now :) 
@JaDogg

hv7214:
@kmehant i see

JaDogg:
Yeah like Mehant said let's keep current model for now.

PR #920, opened 2020-05-17
Now build shows only errors and other logs are redirected

kmehant:
## Description
Travis build is being failed to huge logs from docker builds.
Now we redirect all normal logs and only show when there is an error during the build process.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,9 +17,7 @@ script:
   - docker build --build-arg VCS_REF=`git rev-parse --short HEAD`   --build-arg BUILD_DATE=`date -u +”%Y-%m-%dT%H:%M:%SZ”` -t scoreucsc/bassa-aria2c components/aria2c >/dev/null
 
 after_success:
-  - echo $TRAVIS_BRANCH
-  - echo $TRAVIS_PULL_REQUEST
-  - if [ "$TRAVIS_BRANCH" == "develop" -a "$TRAVIS_PULL_REQUEST" == "false" ]; then
+  - if [ "$TRAVIS_BRANCH" == "develop" ]; then
     docker login -u "$DOCKER_USERNAME" -p "$DOCKER_PASSWORD";
     docker push scoreucsc/bassa-prod-ui;
     docker push scoreucsc/bassa-dev-ui;
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,12 +16,12 @@ script:
   - docker build --build-arg VCS_REF=`git rev-parse --short HEAD`   --build-arg BUILD_DATE=`date -u +”%Y-%m-%dT%H:%M:%SZ”` -t scoreucsc/bassa-dev-server components/core >/dev/null
   - docker build --build-arg VCS_REF=`git rev-parse --short HEAD`   --build-arg BUILD_DATE=`date -u +”%Y-%m-%dT%H:%M:%SZ”` -t scoreucsc/bassa-aria2c components/aria2c >/dev/null
 
-after_success:
-  - if [ "$TRAVIS_BRANCH" == "develop" ]; then
-    docker login -u "$DOCKER_USERNAME" -p "$DOCKER_PASSWORD";
-    docker push scoreucsc/bassa-prod-ui;
-    docker push scoreucsc/bassa-dev-ui;
-    docker push scoreucsc/bassa-prod-server;
-    docker push scoreucsc/bassa-dev-server
-    docker push scoreucsc/bassa-aria2c;
-    fi
+deploy:
+  - provider: script
+    script: deploy.sh
+    on:
+      branch: master
+  - provider: script
+    script: deploy.sh
+    on:
+      branch: develop
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,22 +1,25 @@
 dist: trusty
-language: minimal
+language: generic
 
 sudo: required
 
 services:
   - docker
 
+notifications:
+  webhooks: https://www.travisbuddy.com/
 
-script: 
-  - docker build --build-arg VCS_REF=`git rev-parse --short HEAD`   --build-arg BUILD_DATE=`date -u +”%Y-%m-%dT%H:%M:%SZ”` -t scoreucsc/bassa-prod-ui ui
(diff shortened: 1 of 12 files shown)


kmehant:
> Hey @kmehant,
> Something went wrong with the build.
> 
> TravisCI finished with status `errored`, which means the build failed because of something unrelated to the tests, such as a problem with a dependency or the build process itself.
> 
> [View build log](https://travis-ci.org/scorelab/Bassa/builds/688013746)
> 
> ###### TravisBuddy Request Identifier: 03bf9c90-9831-11ea-b97e-ed27cf305a90

@JaDogg @agentmilindu 
Everything is working fine now, but we need to set docker hub credentials in our Travis for Bassa so that we can push those images and the red cross is due to that only :)

kmehant:
@agentmilindu @JaDogg 

Another choice is to have container builds.

And how about we manually pushing those images to the hub.

deploy.sh
```diff
@@ -0,0 +1,13 @@
+#! /bin/bash
+
+docker build --build-arg VCS_REF=`git rev-parse --short HEAD`   --build-arg BUILD_DATE=`date -u +”%Y-%m-%dT%H:%M:%SZ”` -t scoreucsc/bassa-prod-ui ui >/dev/null
```
JaDogg, line 3: Can you add some comments to this script. This is a bit hard to understand. 😁😁😁

kmehant: Yupp :)

kmehant:
@agentmilindu @JaDogg 
Please review 😁

JaDogg:
@kmehant I've approved this. :)

kmehant:
@agentmilindu I wish Milindu would give his review too

components/core/Dockerfile.prod
```diff
@@ -19,7 +19,7 @@ RUN apt-get update && \
     apt-get clean
 
 # Installing the python modules
-RUN python3 setup.py develop
+RUN ["/bin/bash", "-c", "python3", "setup.py", "develop"]
```
agentmilindu, line 22: Just curious, why do we need this change? Isn't docker doing the same with Shell form?

.travis.yml
```diff
@@ -1,27 +1,27 @@
...
-    docker login -u "$DOCKER_USERNAME" -p "$DOCKER_PASSWORD";
-    docker push scoreucsc/bassa-prod-ui;
-    docker push scoreucsc/bassa-dev-ui;
-    docker push scoreucsc/bassa-prod-server;
-    docker push scoreucsc/bassa-dev-server
-    docker push scoreucsc/bassa-aria2c;
-    fi
+script:
+  - docker build --build-arg VCS_REF=`git rev-parse --short HEAD`   --build-arg BUILD_DATE=`date -u +”%Y-%m-%dT%H:%M:%SZ”` -t scoreucsc/bassa-prod-ui ui >/dev/null
+  - docker build -f ui/Dockerfile.dev --build-arg VCS_REF=`git rev-parse --short HEAD`   --build-arg BUILD_DATE=`date -u +”%Y-%m-%dT%H:%M:%SZ”` -t scoreucsc/bassa-dev-ui ui >/dev/null
```
agentmilindu, line 14: Do we have separate Docker repos as prod/dev? Can't we use docker tags to differentiate Dev and Prod images?

agentmilindu, line 14: Can't we define the date at the beginning to an variable and then reuse it for all the commands? If we have to change it, then we have to change several places. And this does not look tidy.

agentmilindu: Same goes with Git SHA

components/core/Dockerfile.prod
kmehant: @agentmilindu 
True this is weird but I have to say this,

I have my docker run failing on my Mac if I do it with bash 
But it is other way round on Travis where it is failing due to `sh`.
What I feel is it's better to push it manually with sh itself. 
I am not sure why this is happening, do you think of any other workaround.

Also, when I approached a few dockerfile maintainers, and what I learned was they do the docker push part manually always.  
Please let me know your views?

deploy.sh
```diff
@@ -0,0 +1,18 @@
...
+# shows only logs pushed to stderr
+docker build --build-arg VCS_REF=`git rev-parse --short HEAD`   --build-arg BUILD_DATE=`date -u +”%Y-%m-%dT%H:%M:%SZ”` -t scoreucsc/bassa-prod-ui ui >/dev/null
+docker build -f ui/Dockerfile.dev --build-arg VCS_REF=`git rev-parse --short HEAD`   --build-arg BUILD_DATE=`date -u +”%Y-%m-%dT%H:%M:%SZ”` -t scoreucsc/bassa-dev-ui ui >/dev/null
+docker build -f components/core/Dockerfile.prod --build-arg VCS_REF=`git rev-parse --short HEAD`   --build-arg BUILD_DATE=`date -u +”%Y-%m-%dT%H:%M:%SZ”` -t scoreucsc/bassa-prod-server components/core >/dev/null
+docker build --build-arg VCS_REF=`git rev-parse --short HEAD`   --build-arg BUILD_DATE=`date -u +”%Y-%m-%dT%H:%M:%SZ”` -t scoreucsc/bassa-dev-server components/core >/dev/null
+docker build --build-arg VCS_REF=`git rev-parse --short HEAD`   --build-arg BUILD_DATE=`date -u +”%Y-%m-%dT%H:%M:%SZ”` -t scoreucsc/bassa-aria2c components/aria2c >/dev/null
+
+# push both production and development use docker images to dockerhub
+docker login -u "$DOCKER_USERNAME" -p "$DOCKER_PASSWORD"
+docker push scoreucsc/bassa-prod-ui
```
agentmilindu, line 14: Think about tagging the Docker images with release version also. Push to both `latest` and release tags

.travis.yml
kmehant: That is cool will have something like 

these tags
`:prod` for production builds 
`:dev` for development builds 
is that okay?

kmehant: Will do it. :)

deploy.sh
kmehant: Will do this too :)

components/core/Dockerfile.prod
JaDogg: Add a comment and we can keep

kmehant:
reopening this here https://github.com/scorelab/Bassa/pull/921
So closing this :)

PR #890, opened 2020-02-20
Add blueprints to make the routes look more manageable

carpecodeum:
## Description
Added blueprints to make the routes look more manageable and created a view directory to store each type of endpoints seperately.

## Motivation and Context
the system for managing the endpoints was very clustered now it looks better and more manageable.

## How Has This Been Tested?
Tested by performing different activities using different routes.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #927, opened 2020-08-03
Get backend URL from env

kmehant:
Signed-off-by: K mehant <411843@student.nitandhra.ac.in>


kmehant:
Use environment variable `apiUrl` to set Bassa API url

ui/Dockerfile
```diff
@@ -12,8 +12,10 @@ RUN apt-get update && \
     npm install && \
     npm install --global bower gulp-cli && \
     bower --allow-root install
-    
-    
+
+# Add environment variable values to env.js
+# ENV: apiUrl : backend url
+RUN sh env.sh
```
agentmilindu, line 18: This won't work, this will run on the build time not on run time. Users who pull the docker image from the Docker hub won't be able to change the values when running.

kmehant: @agentmilindu Made a workaround, Please look at it.

agentmilindu:
Screenshot from 2020-08-04 15-39-19

Try this method.

See https://hub.docker.com/_/nginx

ui/env-prod.sh
```diff
@@ -0,0 +1,5 @@
+
+pattern='  window.__env.apiUrl.*'
+replacement="  window.__env.apiUrl = '${apiUrl}';"
+
+sed -i -e 's,'"$pattern"','"$replacement"',' /usr/share/nginx/html/env.js
```
JaDogg, line 5: is it not possible to move this to a `.json` file and patch that accordingly or change it accordingly? What is the reason for patching a `.js` file?

kmehant: @JaDogg Actually I used the javascript file which is imported in static HTML file so that whenever the homepage is loaded, the script is run and the variables are as well loaded.

kmehant:
@agentmilindu I have tested it now, it works well having `apiUrl` set for server URL is taking up the changes in run time on reloading the container. 

But all server routes give out 404 because of this PR https://github.com/scorelab/Bassa/pull/890 which does not seem to work with WSGI and supervisord layers well. Works without WSGI and supervisord though.

kmehant:
> Screenshot from 2020-08-04 15-39-19
> 
> Try this method.
> 
> See https://hub.docker.com/_/nginx
> 
> COPY the `env.js.template` to `/etc/nginx/templates` with placeholder

Are you planning to have bassa server also routed through the same nginx 🤔, then how about mounting the nginx config with an another server block for Bassa server?

